PR #2, opened 2020-04-01
[WIP] Monadical's first implementation of python bindings for libzim

Monadical-Inc:
## Progress

- [x] implement main ZimReader and ZimWriter interface via cython
- [x] (partial) finish python packaging setup (`setup.py`, `sdist` & `bdist` setup, `tox`?, setup `README.md` md parsing instead of rst)
- [ ] finish README with more examples, usage, dev instructions, API reference docs, etc.
- [ ] setup CI to build, test, and release automatically
* [x] debug ZimArticle content writes
* [ ] ask about legacy zim format support
* [ ] check libzim tests to reproduce and use their test files
* [ ] make sure all tests are passing (creating, reading & writing)

---

## Goals

- Will become the canonical python-facing API for creating and interacting with ZIM files, used both internally by the scrapers and externally by users consuming ZIM files
- Will start by exposing the core `libzim` API used by `sotoki` and `mwoffliner` (described below), with potential to expand to the rest of the `libzim` functionality
- Can be partly  property-based auto-generated Cython code, based on `libzim`'s `.h` files 
- Should be threadsafe, usable with `async/await`, `multiprocessing`, or `Threading`
- Should be built and tested w/ CI/CD and good documentation

---

#### 1. Article
 
 - [x] ZimArticle constructor (that supports taking binary blobs of arbitrary MIME types)

#### 2. ZimCreator

- [x] ZimCreator constructor (we've already got this working)
- [x] addArticle (in-progress, a little tricky to deal with passing article blob object)
- [x] setMetadata (pretty easy, can just take some key/value pairs)
- [x] finalise (super easy)

#### 3. ZimReader

- [x] ZimReader constructor (we've already got this working)
- [x] destroy (can be a method in python land, or we can hook it up to `__del__`?)
- [x] getCountArticles (we've already got this working)
- [x] getArticleByUrl (super easy, not much work to implement)
- [x] getArticleById (same here)
- [x] suggest (easy to call, but returns an iterator that may need some work to translate)
- [x] search (easy to call, but returns an iterator that may need some work to translate)

Our work on this so far was inspired by Matthew G.'s existing code from here:

---

## Resources

 - https://framagit.org/mgautierfr/pyzim + https://framagit.org/mgautierfr/pyzim/-/blob/master/pyzim/zim_wrapper.pxd
 - https://github.com/pediapress/pyzim
 - https://github.com/jarondl/pyzimmer/blob/master/pyzimmer/zim_writer.py
 - https://github.com/cython/cython/wiki/AutoPxd
 - https://www.youtube.com/watch?v=YReJ3pSnNDo
 - https://github.com/openzim/zim-tools/blob/master/src/zimrecreate.cpp (Article's Abstract Class wrapper)
 - [Paul Ganssle -- Build your Python Extensions with Rust!](https://www.youtube.com/watch?v=1KC43QuJKlE) (also talks about C++)


kelson42:
I'm not sure to understand why with have Docker stuff in this PR. Could you please explain it?

rgaudin:
Thank you very much!

I haven't tested it already but I will. In the mean time, just taking a quick look at it, I have a few questions:

* we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?
* Usage of `finalize()` seems optional? Is that so? Why?
* I think we want only unicode literrals (so no `u"some string"`) as this is py3 only.
* why are there `.pyc` files in the repo?
* Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?
* I dislike the `add_art()` vs `add_article()` methods. Exposed API should be named clearly.
* Maybe `ZimCreator` and `ZimReader` might benefit from some docstrings.
* Naming still, this one is lower-level so it doesn't matter much but the following seems odd:

``` py
creator = pyzim.Creator(True)
creator.zim_creation('hola.zim',True,"eng",2048)
```

Again, this is all from a quick read of the PR ; will test properly. Thanks again 👍

jdcaballerov:
> Thank you very much!
> 
> I haven't tested it already but I will. In the mean time, just taking a quick look at it, I have a few questions:
> 
> * we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?
> * Usage of `finalize()` seems optional? Is that so? Why?
> * I think we want only unicode literrals (so no `u"some string"`) as this is py3 only.
> * why are there `.pyc` files in the repo?
> * Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?
> * I dislike the `add_art()` vs `add_article()` methods. Exposed API should be named clearly.
> * Maybe `ZimCreator` and `ZimReader` might benefit from some docstrings.
> * Naming still, this one is lower-level so it doesn't matter much but the following seems odd:
> 
> ```python
> creator = pyzim.Creator(True)
> creator.zim_creation('hola.zim',True,"eng",2048)
> ```
> 
> Again, this is all from a quick read of the PR ; will test properly. Thanks again 👍



> I'm not sure to understand why with have Docker stuff in this PR. Could you please explain it?

@kelson42  To compile the python extension libzim is needed, since some people might use Mac and libzim is hard to get to compile on a Mac I set up a docker container that can be used for development and testing.

jdcaballerov:
> Thank you very much!
> 
> I haven't tested it already but I will. In the mean time, just taking a quick look at it, I have a few questions:

@rgaudin 

I advice to use the latest HEAD of the branch python3 where I cleaned the code a lot and already addressed some bugs and comments 

> 
> * we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?

Easily changed, up to you, just bear in mind there are some python naming restrictions for packages.

> * Usage of `finalize()` seems optional? Is that so? Why?

No, it's needed it might be an error at the example. finalise calls 
`creator->finishZimCreation();`

> * I think we want only unicode literrals (so no `u"some string"`) as this is py3 only.

Yes, initially I started developing in python 2.7 but all was moved to python3. It might be a mistake.

> * why are there `.pyc` files in the repo?

Mistake.

> * Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?

Just included for tests, perhaps the ones included with libzim are more appropriate.

> * I dislike the `add_art()` vs `add_article()` methods. Exposed API should be named clearly.

add_art was a test version I used while debugging. Will not be included.

> * Maybe `ZimCreator` and `ZimReader` might benefit from some docstrings.

documentation and docstrings are pending once you are ok with the API.

> * Naming still, this one is lower-level so it doesn't matter much but the following seems odd:
> 
> ```python
> creator = pyzim.Creator(True)
> creator.zim_creation('hola.zim',True,"eng",2048)
> ```

It's was changed, the example file included was outdated.

> 
> Again, this is all from a quick read of the PR ; will test properly. Thanks again 👍



> Thank you very much!
> 
> I haven't tested it already but I will. In the mean time, just taking a quick look at it, I have a few questions:
> 
> * we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?
> * Usage of `finalize()` seems optional? Is that so? Why?
> * I think we want only unicode literrals (so no `u"some string"`) as this is py3 only.
> * why are there `.pyc` files in the repo?
> * Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?
> * I dislike the `add_art()` vs `add_article()` methods. Exposed API should be named clearly.
> * Maybe `ZimCreator` and `ZimReader` might benefit from some docstrings.
> * Naming still, this one is lower-level so it doesn't matter much but the following seems odd:
> 
> ```python
> creator = pyzim.Creator(True)
> creator.zim_creation('hola.zim',True,"eng",2048)
> ```
> 
> Again, this is all from a quick read of the PR ; will test properly. Thanks again 👍

mgautierfr:
I haven't made a review of your code, but just base on previous comments :

> @kelson42 To compile the python extension libzim is needed, since some people might use Mac and libzim is hard to get to compile on a Mac I set up a docker container that can be used for development and testing.

To use the python extension you also need libzim. And you wouldn't set a docker container (if you can) to use a library. Why compile libzim is hard on Mac ? kiwix-build is here for that.

> No, it's needed it might be an error at the example. finalise calls
creator->finishZimCreation();

I would prefer to keep the method `finishZimCreation()` alone. And memory management shouldn't be exposed to the user. The cpp instance must be deleted when the gc deletes the python object.

> Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?

That is a real question that goes beyond this project. libzim also have some test zim and I somehow dislike this.
We could store them in a server, but then the test process would have to download it each time (or implement some kind of cache).
Anyway, those 25MB are already in the repository. Remove them will not reduce the repository size.

> creator.zim_creation('hola.zim',True,"eng",2048)

I agree with @rgaudin comment and I would add that function with to many arguments is difficult to use. A better way would be to force the use of
keyword args : `...creation('hola.zim', lang="eng", clusterSize=2048, verbose=True)`

rgaudin:
Thanks for the update

FYI, when trying to build on macOS (I have libzim installed), I get an error

```
In file included from pyzim/pyzim.cpp:651:
pyzim/wrappers.cpp:17:7: error: exception specification of overriding function is more lax than base version
class ZimArticle : public zim::writer::Article
```

Following are a few comments on the python usage side.

* better `__repr__` of objects like `ZimReader` (filename?), `ZimArticle` (title ? url?), `ZimCreator` (filename? lang?)
* what's the use of `article.get_article_properties()` ?
* shouldn't we get a generator on `reader.search()` ? Do the libzim returns a list as-is? It's frequent to have thousands of results on large ZIM files.

``` py
reader.search("fdisk")
> ['A/Plan_9', 'A/GNU/Linux', 'A/OpenBSD']
```

* already mentioned (upstream bug?) but

``` py
art.is_redirect
> False
art.redirect_url
> ck_policyE2EEfavicon
art.redirect_longurl
> A/ck_policyE2EEfavicon
```

* `constructor docstrings are not usable for `ZimCreator`, `ZimReader` and `ZimArticle`.
* probably upstream limitation but requiring `main_page` on `ZimCreator` creation seems weird. You allow to set it to blank but can't change it afterwards?
* you chose to Pascalize meta key names. The [documentation](https://wiki.openzim.org/wiki/Metadata) doesn't require it. It just happens to be the case (ahah) for currently defined ones but as stated there, it's extendable. @kelson42 ?
https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L576
* related, it seems that `get_metadata` only returns those predefined keys while `ZimCreator` allows setting any. Confusing.
https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L308
https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L559
* side-note, I know this is cython but I think redefining builtins like `id` is bad practice and should be avoided.
https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L317 
* on code formatting, for the pure-python files, (examples, tests), could you please use [black](https://pypi.org/project/black/) formatting? It'd help to use it (as can be) on the cython file as well.
* adding a new article to an already finalized creator causes a segfault.
* crashes when requesting by id on an invalid Id

``` py
reader.get_article_by_id(7).title
terminate called after throwing an instance of 'zim::ZimFileFormatError'
  what():  article index out of range
Aborted
```

* I see `finalise()` which is british english. I believe we use US English (`finalize()`). How's it in the libzim? @kelson42 ?
* we have no access to params passed to `ZimCreator()` : `main_page`, `index_language`, etc.
* there's an unsafe use of memory. if you use the same variable for different articles, the articles count of the zim there are added to gets messed-up

``` py
creator = pyzim.ZimCreator("z1.zim")
article = pyzim.ZimArticle(url="1", title="Page 1", content="<html><head></head><body><h1>Welcome to page 1</h1></body></html>")
creator.add_article(article)
creator.finalise()
# T:0; A:4; RA:0; CA:4; UA:0; FA:0; IA:4; C:0; CC:0; UC:0; WC:3
# T:0; Generate cluster offsets
# T:0; ResolveRedirectIndexes
# Resolve redirect
# T:0; Set article indexes
# set index
# T:0; Resolve mimetype
# T:0; create title index
# T:0; 6 title index created
# T:0; 2 clusters created
# T:0; fill header
# T:0; write zimfile :
# T:0;  write header
# T:0;  write mimetype list
# T:0;  write url prt list
# T:0;  write title index
# T:0;  write directory entries
# T:0;  write cluster offset list
# T:0;  write cluster data
# T:0;  write checksum
# T:0; finish
reader = pyzim.ZimReader("z1.zim")
reader.get_article_count()  # 6
reader.get_namespaces_count("A")  # 1 (OK)

creator2 = pyzim.ZimCreator("z2.zim")
article = pyzim.ZimArticle(url="2", title="Page 2", content="<html><head></head><body><h1>Welcome to page 2</h1></body></html>")
creator2.add_article(article)
creator2.finalise()
# T:0; A:4; RA:0; CA:4; UA:0; FA:0; IA:4; C:0; CC:0; UC:0; WC:1
# T:0; Generate cluster offsets
# T:0; ResolveRedirectIndexes
# Resolve redirect
# T:0; Set article indexes
# set index
# T:0; Resolve mimetype
# T:0; create title index
# T:0; 6 title index created
# T:0; 2 clusters created
# T:0; fill header
# T:0; write zimfile :
# T:0;  write header
# T:0;  write mimetype list
# T:0;  write url prt list
# T:0;  write title index
# T:0;  write directory entries
# T:0;  write cluster offset list
# T:0;  write cluster data
# T:0;  write checksum
# T:0; finish
reader2 = pyzim.ZimReader("z2.zim")
reader2.get_article_count()  # 6
reader2.get_namespaces_count("A")  # 1 (OK)
```

Now add both to kiwix-serve and you'll see that z1 has `1` article while z2 has `2` articles. If we were to reuse that same `article` many times, we'd the the counter increase by that number.
* last but not least, `ZimArticle` only accepts content as `str`. How am I suppose to add binary data to the ZIM ? Are we not there yet or is that a mistake ?

Thanks again for the effort, it's quite pleasant to use already.

jdcaballerov:
> Thanks for the update
> 
> FYI, when trying to build on macOS (I have libzim installed), I get an error
> 
> ```
> In file included from pyzim/pyzim.cpp:651:
> pyzim/wrappers.cpp:17:7: error: exception specification of overriding function is more lax than base version
> class ZimArticle : public zim::writer::Article
> ```

Need to check this. Whats the compiler on your Mac ? I guess is related to the virtual destructor or some missing virtual.

> 
> Following are a few comments on the python usage side.
> 
> * better `__repr__` of objects like `ZimReader` (filename?), `ZimArticle` (title ? url?), `ZimCreator` (filename? lang?)

I'll add what you consider is better.

> * what's the use of `article.get_article_properties()` ?

It was left after a refactor, will be deleted.

> * shouldn't we get a generator on `reader.search()` ? Do the libzim returns a list as-is? It's frequent to have thousands of results on large ZIM files.
> 
> ```python
> reader.search("fdisk")
> > ['A/Plan_9', 'A/GNU/Linux', 'A/OpenBSD']
> ```

I did the same as node-libzim getting 10 results and passing back a string vector. File.h returns Search class unique pointers.

```
 std::unique_ptr<Search> search(const std::string& query, int start, int end) const;
 std::unique_ptr<Search> suggestions(const std::string& query, int start, int end) const;
```

> 
> * already mentioned (upstream bug?) but
> 
> ```python
> art.is_redirect
> > False
> art.redirect_url
> > ck_policyE2EEfavicon
> art.redirect_longurl
> > A/ck_policyE2EEfavicon
> ```

I can restrict (or raise an exception) getting those in Cython code if it's not a redirect if you agree.

> 
> * `constructor docstrings are not usable for `ZimCreator`, `ZimReader`and`ZimArticle`.

Need to be included.

> * probably upstream limitation but requiring `main_page` on `ZimCreator` creation seems weird. You allow to set it to blank but can't change it afterwards?

I used the same construct as node-libzim that fixes it from creation.

`ZimCreator` must implement a function getMainPage so that when `finishZimCreation()` (C++) is called the header is filled with it `fillHeader()`. 

With the current design I can overload the wrapper function and add a setter if needed.

> * you chose to Pascalize meta key names. The [documentation](https://wiki.openzim.org/wiki/Metadata) doesn't require it. It just happens to be the case (ahah) for currently defined ones but as stated there, it's extendable. @kelson42 ?

:)

>   https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L576
> * related, it seems that `get_metadata` only returns those predefined keys while `ZimCreator` allows setting any. Confusing.

Yes it's correct, maybe we can put there only the mandatory metadata. Any metadata can be accessed getting an article by url

```
date_metadata = zim_reader.get_article("M/Date")
counter_metadata = zim_reader.get_article("M/Counter")
```

>   https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L308
>   
>   https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L559
> * side-note, I know this is cython but I think redefining builtins like `id` is bad practice and should be avoided.

Can be changed.

>   https://github.com/openzim/python-libzim/blob/0e9c65a636349cdc4caf746b7a72e3c7fad748b2/pyzim/pyzim.pyx#L317
> * on code formatting, for the pure-python files, (examples, tests), could you please use [black](https://pypi.org/project/black/) formatting? It'd help to use it (as can be) on the cython file as well.

Ok.

> * adding a new article to an already finalized creator causes a segfault.

Will add a safeguard as it's done in finalised() .

> * crashes when requesting by id on an invalid Id
> 
> ```python
> reader.get_article_by_id(7).title
> terminate called after throwing an instance of 'zim::ZimFileFormatError'
>   what():  article index out of range
> Aborted
> ```

Upstream. I'll try to capture the exception in Cython.

```c++
 Article File::getArticle(article_index_type idx) const
  {
    if (idx >= article_index_type(impl->getCountArticles()))
      throw ZimFileFormatError("article index out of range");
    return Article(impl, idx);
  }
```

> 
> * I see `finalise()` which is british english. I believe we use US English (`finalize()`). How's it in the libzim? @kelson42 ?

libzim uses finishZimCreation(),  finalise is used in node-libzim.

> * we have no access to params passed to `ZimCreator()` : `main_page`, `index_language`, etc.

I can add properties if you agree on.

> * there's an unsafe use of memory. if you use the same variable for different articles, the articles count of the zim there are added to gets messed-up
> 
> ```python
> creator = pyzim.ZimCreator("z1.zim")
> article = pyzim.ZimArticle(url="1", title="Page 1", content="<html><head></head><body><h1>Welcome to page 1</h1></body></html>")
> creator.add_article(article)
> creator.finalise()
> # T:0; A:4; RA:0; CA:4; UA:0; FA:0; IA:4; C:0; CC:0; UC:0; WC:3
> # T:0; Generate cluster offsets
> # T:0; ResolveRedirectIndexes
> # Resolve redirect
> # T:0; Set article indexes
> # set index
> # T:0; Resolve mimetype
> # T:0; create title index
> # T:0; 6 title index created
> # T:0; 2 clusters created
> # T:0; fill header
> # T:0; write zimfile :
> # T:0;  write header
> # T:0;  write mimetype list
> # T:0;  write url prt list
> # T:0;  write title index
> # T:0;  write directory entries
> # T:0;  write cluster offset list
> # T:0;  write cluster data
> # T:0;  write checksum
> # T:0; finish
> reader = pyzim.ZimReader("z1.zim")
> reader.get_article_count()  # 6
> reader.get_namespaces_count("A")  # 1 (OK)
> 
> creator2 = pyzim.ZimCreator("z2.zim")
> article = pyzim.ZimArticle(url="2", title="Page 2", content="<html><head></head><body><h1>Welcome to page 2</h1></body></html>")
> creator2.add_article(article)
> creator2.finalise()
> # T:0; A:4; RA:0; CA:4; UA:0; FA:0; IA:4; C:0; CC:0; UC:0; WC:1
> # T:0; Generate cluster offsets
> # T:0; ResolveRedirectIndexes
> # Resolve redirect
> # T:0; Set article indexes
> # set index
> # T:0; Resolve mimetype
> # T:0; create title index
> # T:0; 6 title index created
> # T:0; 2 clusters created
> # T:0; fill header
> # T:0; write zimfile :
> # T:0;  write header
> # T:0;  write mimetype list
> # T:0;  write url prt list
> # T:0;  write title index
> # T:0;  write directory entries
> # T:0;  write cluster offset list
> # T:0;  write cluster data
> # T:0;  write checksum
> # T:0; finish
> reader2 = pyzim.ZimReader("z2.zim")
> reader2.get_article_count()  # 6
> reader2.get_namespaces_count("A")  # 1 (OK)
> ```
> 
> Now add both to kiwix-serve and you'll see that z1 has `1` article while z2 has `2` articles. If we were to reuse that same `article` many times, we'd the the counter increase by that number.
> 

I don't understand completely what's kiwix-serve? can you give me the whole example ?

> * last but not least, `ZimArticle` only accepts content as `str`. How am I suppose to add binary data to the ZIM ? Are we not there yet or is that a mistake ?

For the first version I was using auto conversion from Cython. Can be changed to accept bytes.

> 
> Thanks again for the effort, it's quite pleasant to use already.

rgaudin:
> > FYI, when trying to build on macOS (I have libzim installed), I get an error
>
> Need to check this. Whats the compiler on your Mac ? I guess is related to the virtual destructor or some missing virtual.

`clang` called as `gcc` (default macOS setup)
 
> > * better `__repr__` of objects like `ZimReader` (filename?), `ZimArticle` (title ? url?), `ZimCreator` (filename? lang?)
>
> I'll add what you consider is better.

OK, please use the following for now:

``` py
# ZimArticle
def __repr__(self):
        return f"{self.__class__.__name__}(url={self.longurl}, title=)"

# ZimReader & ZimCreator (requires the filename getter)
def __repr__(self):
        return f"{self.__class__.__name__}(filename={self.filename})"
```

> > * shouldn't we get a generator on `reader.search()` ? Do the libzim returns a list as-is? It's frequent to have thousands of results on large ZIM files.
>
> I did the same as node-libzim getting 10 results and passing back a string vector. File.h returns Search class unique pointers.
> ```
>  std::unique_ptr<Search> search(const std::string& query, int start, int end) const;
>  std::unique_ptr<Search> suggestions(const std::string& query, int start, int end) const;
> ```

Ah ok I see. that's unfortunate. I didn't realized you hardcoded that. We should definitely be able to set those. But we'd need the number of results (from `get_matches_estimated()`) as well. How do you plan to address that?

>> * already mentioned (upstream bug?) but
>
> I can restrict (or raise an exception) getting those in Cython code if it's not a redirect if you agree.

@mgautierfr should answer that. I'd prefer this situation not to happen but ideally that should be fixed upstream.
 
> > * probably upstream limitation but requiring `main_page` on `ZimCreator` creation seems weird. You allow to set it to blank but can't change it afterwards?
>
> I used the same construct as node-libzim that fixes it from creation.
> `ZimCreator` must implement a function getMainPage so that when `finishZimCreation()` (C++) is called the header is filled with it `fillHeader()`.
> With the current design I can overload the wrapper function and add a setter if needed.

I think it makes sense, yes.

> > * you chose to Pascalize meta key names. The [documentation](https://wiki.openzim.org/wiki/Metadata) doesn't require it. It just happens to be the case (ahah) for currently defined ones but as stated there, it's extendable. @kelson42 ?
> > * related, it seems that `get_metadata` only returns those predefined keys while `ZimCreator` allows setting any. Confusing.
>
> Yes it's correct, maybe we can put there only the mandatory metadata. Any metadata can be accessed getting an article by url

Yes, let's harmonize to only manage _official ones_ this way and have the article mechanism for the rest ; if @kelson42  and @mgautierfr  agrees.
 
> > * crashes when requesting by id on an invalid Id
>
> Upstream. I'll try to capture the exception in Cython.

Yeah we definitely not want the runtime to blow-up on mistakes like this one. Other similar crash scenario, creating a `ZimCreator` for which the libzim temp folder exists (`xxx.tmp`) ; or a `ZimReader` on a non-existing file…

> 
> ```c++
>  Article File::getArticle(article_index_type idx) const
>   {
>     if (idx >= article_index_type(impl->getCountArticles()))
>       throw ZimFileFormatError("article index out of range");
>     return Article(impl, idx);
>   }
> ```

That would work but I wonder if there would be a way to catch the actual cpp exception and convert it to a python one. That would be safer than overcoming all the possible incidents from the wrapper.

On that `z.tmp` folder, it is created on `ZimCreator("z.zim")` along with `z.tmp.idx.tmp` and `z.tmp_title.idx.tmp` but if the creator is not finalized, it stays on disk. Should we remove it on garbage collection?

> > * I see `finalise()` which is british english. I believe we use US English (`finalize()`). How's it in the libzim? @kelson42 ?
>
> libzim uses finishZimCreation(), finalise is used in node-libzim.

Ah! well that explains it as @ISNIT0 is British.  Let's switch it to `finalize` while it's a minor change.

> > * we have no access to params passed to `ZimCreator()` : `main_page`, `index_language`, etc.
>
> I can add properties if you agree on.

Yes, please.

> > * there's an unsafe use of memory. if you use the same variable for different articles, the articles count of the zim there are added to gets messed-up
> > Now add both to kiwix-serve and you'll see that z1 has `1` article while z2 has `2` articles. If we were to reuse that same `article` many times, we'd the the counter increase by that number.
> 
> I don't understand completely what's kiwix-serve? can you give me the whole example ?

[kiwix-serve](https://github.com/kiwix/kiwix-tools/) is a web server serving ZIM files that uses the [kiwix-lib](https://github.com/kiwix/kiwix-lib/) which itself used the libzim. On its homepage, there's a counter showing the number of html articles. See [reader.cpp](https://github.com/kiwix/kiwix-lib/blob/12a066034202aaa41c3deeab6e9f020a725f39ef/src/reader.cpp#L131)

``` cpp
/* Get the count of articles which can be indexed/displayed */
unsigned int Reader::getArticleCount() const
{
  std::map<const std::string, unsigned int> counterMap
      = this->parseCounterMetadata();
  unsigned int counter = 0;

  if (counterMap.empty()) {
    counter = this->nsACount;
  } else {
    auto it = counterMap.find("text/html");
    if (it != counterMap.end()) {
      counter = it->second;
    }
  }

  return counter;
}
```
Screen Shot 2020-04-09 at 16 02 49

You can find [binary here](https://download.kiwix.org/release/kiwix-tools/) or use docker image `kiwix/kiwix-serve:latest`. Used `kiwix-serve -p 9999 ~/src/python-libzim/z*.zim`

> > * last but not least, `ZimArticle` only accepts content as `str`. How am I suppose to add binary data to the ZIM ? Are we not there yet or is that a mistake ?
>
> For the first version I was using auto conversion from Cython. Can be changed to accept bytes.

Well we definitely need to be able to set bytes. If we could also set/get text without having to manually encode/decode, that would be great.

jdcaballerov:
> > > FYI, when trying to build on macOS (I have libzim installed), I get an error
> > 
> > 
> > Need to check this. Whats the compiler on your Mac ? I guess is related to the virtual destructor or some missing virtual.
> 
> `clang` called as `gcc` (default macOS setup)
> 
> > > * better `__repr__` of objects like `ZimReader` (filename?), `ZimArticle` (title ? url?), `ZimCreator` (filename? lang?)

Issue #4 

> > 
> > 
> > I'll add what you consider is better.
> 
> OK, please use the following for now:
> 
> ```python
> # ZimArticle
> def __repr__(self):
>         return f"{self.__class__.__name__}(url={self.longurl}, title=)"
> 
> # ZimReader & ZimCreator (requires the filename getter)
> def __repr__(self):
>         return f"{self.__class__.__name__}(filename={self.filename})"
> ```

Done ✅

> 
> > > * shouldn't we get a generator on `reader.search()` ? Do the libzim returns a list as-is? It's frequent to have thousands of results on large ZIM files.
> > 
> > 
> > I did the same as node-libzim getting 10 results and passing back a string vector. File.h returns Search class unique pointers.
> > ```
> >  std::unique_ptr<Search> search(const std::string& query, int start, int end) const;
> >  std::unique_ptr<Search> suggestions(const std::string& query, int start, int end) const;
> > ```
> 
> Ah ok I see. that's unfortunate. I didn't realized you hardcoded that. We should definitely be able to set those. But we'd need the number of results (from `get_matches_estimated()`) as well. How do you plan to address that?
> 
@rgaudin I'll first need to understand what you want to do with Search. 

I might get a Search pointer from File.h search function, once in python search function I could use the iterator and yield. But please let me know  What's the use case of `get_matches_estimated() ` ? 

Issue #5  


> > > * already mentioned (upstream bug?) but
> > 
> > 
> > I can restrict (or raise an exception) getting those in Cython code if it's not a redirect if you agree.
> 
> @mgautierfr should answer that. I'd prefer this situation not to happen but ideally that should be fixed upstream.

I'll recheck tomorrow how redirects are used.

> > > * probably upstream limitation but requiring `main_page` on `ZimCreator` creation seems weird. You allow to set it to blank but can't change it afterwards?
> > 
> > 
> > I used the same construct as node-libzim that fixes it from creation.
> > `ZimCreator` must implement a function getMainPage so that when `finishZimCreation()` (C++) is called the header is filled with it `fillHeader()`.
> > With the current design I can overload the wrapper function and add a setter if needed.
> 
> I think it makes sense, yes.

Issue #6 

> 
> > > * you chose to Pascalize meta key names. The [documentation](https://wiki.openzim.org/wiki/Metadata) doesn't require it. It just happens to be the case (ahah) for currently defined ones but as stated there, it's extendable. @kelson42 ?
> > > * related, it seems that `get_metadata` only returns those predefined keys while `ZimCreator` allows setting any. Confusing.
> > 
> > 
> > Yes it's correct, maybe we can put there only the mandatory metadata. Any metadata can be accessed getting an article by url
> 
> Yes, let's harmonize to only manage _official ones_ this way and have the article mechanism for the rest ; if @kelson42 and @mgautierfr agrees.

Issue #7 

> 
> > > * crashes when requesting by id on an invalid Id
> > 
> > 
> > Upstream. I'll try to capture the exception in Cython.
> 
> Yeah we definitely not want the runtime to blow-up on mistakes like this one. Other similar crash scenario, creating a `ZimCreator` for which the libzim temp folder exists (`xxx.tmp`) ; or a `ZimReader` on a non-existing file…
> 
> > ```c++
> >  Article File::getArticle(article_index_type idx) const
> >   {
> >     if (idx >= article_index_type(impl->getCountArticles()))
> >       throw ZimFileFormatError("article index out of range");
> >     return Article(impl, idx);
> >   }
> > ```
> 
> That would work but I wonder if there would be a way to catch the actual cpp exception and convert it to a python one. That would be safer than overcoming all the possible incidents from the wrapper.
> 
> On that `z.tmp` folder, it is created on `ZimCreator("z.zim")` along with `z.tmp.idx.tmp` and `z.tmp_title.idx.tmp` but if the creator is not finalized, it stays on disk. Should we remove it on garbage collection?
> 

Done ✅ All exceptions will be [catched and translated to Cython](https://github.com/openzim/python-libzim/issues/4#issue-597538722)
Screen Shot 2020-04-09 at 16 16 45


> > > * I see `finalise()` which is british english. I believe we use US English (`finalize()`). How's it in the libzim? @kelson42 ?
> > 
> > 
> > libzim uses finishZimCreation(), finalise is used in node-libzim.
> 
> Ah! well that explains it as @ISNIT0 is British. Let's switch it to `finalize` while it's a minor change.
> 
Done ✅

> > > * we have no access to params passed to `ZimCreator()` : `main_page`, `index_language`, etc.
> > 
> > 
> > I can add properties if you agree on.
> 
> Yes, please.

Done ✅

> 
> > > * there's an unsafe use of memory. if you use the same variable for different articles, the articles count of the zim there are added to gets messed-up
> > >   Now add both to kiwix-serve and you'll see that z1 has `1` article while z2 has `2` articles. If we were to reuse that same `article` many times, we'd the the counter increase by that number.
> > 
> > 
> > I don't understand completely what's kiwix-serve? can you give me the whole example ?
> 
> [kiwix-serve](https://github.com/kiwix/kiwix-tools/) is a web server serving ZIM files that uses the [kiwix-lib](https://github.com/kiwix/kiwix-lib/) which itself used the libzim. On its homepage, there's a counter showing the number of html articles. See [reader.cpp](https://github.com/kiwix/kiwix-lib/blob/12a066034202aaa41c3deeab6e9f020a725f39ef/src/reader.cpp#L131)
> 
> ```c++
> /* Get the count of articles which can be indexed/displayed */
> unsigned int Reader::getArticleCount() const
> {
>   std::map<const std::string, unsigned int> counterMap
>       = this->parseCounterMetadata();
>   unsigned int counter = 0;
> 
>   if (counterMap.empty()) {
>     counter = this->nsACount;
>   } else {
>     auto it = counterMap.find("text/html");
>     if (it != counterMap.end()) {
>       counter = it->second;
>     }
>   }
> 
>   return counter;
> }
> ```
> 
> Screen Shot 2020-04-09 at 16 02 49
> 
> You can find [binary here](https://download.kiwix.org/release/kiwix-tools/) or use docker image `kiwix/kiwix-serve:latest`. Used `kiwix-serve -p 9999 ~/src/python-libzim/z*.zim`
> 

issue #8 

> > > * last but not least, `ZimArticle` only accepts content as `str`. How am I suppose to add binary data to the ZIM ? Are we not there yet or is that a mistake ?
> > 
> > 
> > For the first version I was using auto conversion from Cython. Can be changed to accept bytes.
> 
> Well we definitely need to be able to set bytes. If we could also set/get text without having to manually encode/decode, that would be great.

Done ✅

The constructor will not type content and leave bytes untouched, str are encoded/decoded to utf-8:

```python
    def __cinit__(self, url="", content="", namespace= "A", mimetype= "text/html", title="", redirect_article_url= "",filename="", should_index=True ):
```
Then will encode if it's a python str or pass directly if bytes are inputed:

```python3
bytes_content =b''
        if isinstance(content, str):
            bytes_content = content.encode('UTF-8')
        else:
            bytes_content = content
```

The getter and setter are also adjusted

```python3
@property
    def content(self):
        """Get the article's content"""
        data = self.c_zim_article.content 
        try:
            return data.decode('UTF-8')
        except UnicodeDecodeError:
            return data

    @content.setter
    def content(self, new_content):
        """Set the article's content"""
        if isinstance(new_content,str):
            self.c_zim_article.content = new_content.encode('UTF-8') 
        else:
            self.c_zim_article.content = new_content 
```

setup.py
```diff
@@ -0,0 +1,28 @@
...
+    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+
+setup(
+    name = "python-libzim",
+    version = "0.0.1",
+    author = "Monadical SAS",
+    author_email = "hello@monadical.com",
+    description = ("A python-facing API for creating and interacting with ZIM files"),
+    license = "GPLv3",
+    long_description=read('README.md'),
```
pirate, line 20:
```suggestion
    long_description=read('README.md'),
    long_description_content_type='text/markdown',
```

setup.py
```diff
@@ -0,0 +1,28 @@
...
+# README file and 2) it's easier to type in the README file than to put a raw
+# string in below ...
+def read(fname):
+    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+
+setup(
+    name = "python-libzim",
+    version = "0.0.1",
+    author = "Monadical SAS",
+    author_email = "hello@monadical.com",
```
pirate, line 17:
```suggestion
    author_email = "jdc@monadical.com",
```
😉

kelson42:
>     * we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?

To avoid any confusion, we should call it from the name of the repo I have created.

kelson42:
> Is that wise to have ZIM files in the repo for tests? It's 25MB already. @kelson42 ?

This is a bit heavy indeed. Why so big?

rgaudin:
> >     * we should probably have discussed that from day1 but are we all OK with `pyzim` ? @mgautierfr does that matches other bindings and concurrent projects?
> 
> 
> 
> To avoid any confusion, we should call it from the name of the repo I have created.

this is not a valid python module name. we could have:

- `python_libzim` not recommended to use python in lib name and it’s long.
- **`pylibzim`** sounds a lot better and matches the repo somewhat.
- `libzim` seems obvious and definitely matches the repo (minus the platform). has it been used already ?

I like `pyzim` but there are already 2 other projects with that name I think.

kelson42:
@rgaudin in pip, it should be `libzim`

rgaudin:
> @rgaudin in pypi, it should be `libzim`

then that’s settled. thanks

pirate:
Ok, the biggest remaining things to do are:

- [ ] finalize package naming and make it consistent everywhere
- [ ] fix last remaining tickets #5 #7 and merge PRs
- [ ] setup CI/CD for black, tests, build, and deploy to pypi #9
- [ ] start work to hook up sotoki in parallel: https://github.com/openzim/sotoki/issues/130

jdcaballerov:
> No need for a deepcopy here. If `self._metadata_keys` is a dict as you define it `self._metadata = dict(self._metadata_keys)` is enough.
> If it is a list of keys (as it should be) : `self._metadata = {k:None for k in METADATA_KEYS}`

Agree.

pirate:
Since all efforts are now focused on the blend approach in #3 , we're closing this PR for now.

mgautierfr:
Not everything is to throw here.
The reading part (except the article) is pretty good. You can keep it and just rewrite the read article wrapper is a simpler form.

jdcaballerov:
> Not everything is to throw here.
> The reading part (except the article) is pretty good. You can keep it and just rewrite the read article wrapper is a simpler form.

That's the plan. To reuse most of the reader code perhaps naming as ZimFileReader to avoid confusing with article definitions, but without that particular article implementation. Just simple wrappers for File.h and Article.h (zim::)

PR #3, opened 2020-04-07
Monadical's implementation of python bindings for libzim using the "class blend" approach

pirate:
Continuation of our work here https://github.com/openzim/python-libzim/pull/2#issue-397272514

This PR uses the new class blend approach as discussed between @mgautierfr and @jdcaballerov .

I separated them so that the new work is visible independently, however if you'd prefer we work in a single PR, let me know and I will delete this one and push the work to the other PR.


kelson42:
We have a brand new release of libzim to be used here http://download.openzim.org/release/libzim/libzim_linux-x86_64-6.1.0.tar.gz

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
     def can_write(self):
-        """Get if the article is valid for writing"""
-        # An article must have at least non empty url to be writable
-        # Content can be empty if article is a redirect
-        if self.longurl and (self.content or self.is_redirect):
-            self._can_write = True
-        else:
-            self._can_write = False
-        return self._can_write
+        raise NotImplementedError
```
mgautierfr, line 76: This should not be necessary has this is the type of the Article who define if we can use it to write a zim file.

(It IS not necessary, but maybe for technical reason as handling python ducktyping you may want a marker. But it should not be to the (user) implementation to change this).

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
+            raise 
+        else:
+            error[0] = 0
+            blob = func()
+            return dereference(blob.c_blob) 
 
-    def __dealloc__(self):
-        if self.c_file != NULL:
-            del self.c_file
+cdef public api:
```
mgautierfr, line 109: Why a second `cdef public api:` ?

Those functions are mainly utility function used by our c++ code.
Even if not necessary, it may be a good thing to move them in a different `pyx` file.

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
+            error[0] = 0
+            return <uint64_t> func()
 
-        # TODO Ask about the old format, check libzim for tests
-        # Handle old zim where header has no mainPage.
-        # (We need to get first article in the zim)
-        article = self.c_file.getArticle(<int> 0)
-        if article.good():
-            return article.getLongUrl().decode("UTF-8", "strict")
+class ZimTestArticle(ZimArticle):
```
mgautierfr, line 136: A `ZimTestArticle` should not be in the library definition.
At best in a example implementation (but in a different directory that the package src)

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
 
 #########################
 #       ZimCreator      #
 #########################
 
+
 cdef class ZimCreator:
-    cdef zim.ZimCreator *c_creator
+    cdef ZimCreatorWrapper *c_creator
     cdef object _finalised
```
mgautierfr, line 181: May we use a bool instead of a object ?

pyzim/pyzim.pyx
```diff
@@ -410,10 +214,26 @@ cdef class ZimCreator:
...
+    def _update_article_counter(self, ZimArticle article):
+        # default dict update
+        self._article_counter[article.mimetype] += 1
+
+    def add_article(self, ZimArticle article not None):
+        if not article.can_write:
+            raise RuntimeError("Article is not good for writing")
+
+        # Make a shared pointer to ZimArticleWrapper from the ZimArticle object (dereference internal c_article)
+        cdef shared_ptr[ZimArticleWrapper] art = make_shared[ZimArticleWrapper](dereference(article.c_article));
```
mgautierfr, line 226: I'm not sure about the dereference.
If the ZimArticleWrapper is a wrapper around a python article, we probably don't want to pass it a `*(article.c_article)`

pyzim/pyzim.pyx
```diff
@@ -410,10 +214,26 @@ cdef class ZimCreator:
...
+        try:
+            self.c_creator.addArticle(art)
+        else:
+            if not article.is_redirect:
+                self._update_article_counter(article)
```
mgautierfr, line 231: A `try` without an `except` is mostly useless.
It is even not valid python.

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
-        c_zim_art = new zim.ZimArticle(ord(namespace),                     # Namespace
-                                       url.encode('UTF-8'),                # url
-                                       title.encode('UTF-8'),              # title
-                                       mimetype.encode('UTF-8'),           # mimeType
-                                       redirect_article_url.encode('UTF-8'),# redirectUrl
-                                       should_index,                                # shouldIndex
-                                       content.encode('UTF-8'))
-        self.__setup(c_zim_art)
+    def __init__(self, bytes content):
+        self.c_blob = new Blob(<char *> content, len(content))
```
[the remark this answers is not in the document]
jdcaballerov: > I'm not sure to understand the status of this PR.

This is a WIP using the cython blend approach. Currently I am addressing issues from the first version #2 . That will help me a lot defining the API and understanding the use cases. The idea es to keep the same API and other than using the article subclasses for each task.

> What we discuss about is to have two different wrapper for the articles but I see only one.

Locally it's more advanced (unpushed) . I do have a read article as follows:

```python3
cdef class ZimReadArticle(ZimArticle):
    cdef Article* c_read_article

    def __init__(self):
        ZimArticle.__init__(self)

    cdef __setup(self, Article *art):
        """Assigns an internal pointer to the wrapped C++ article object.
        
        Parameters
        ----------
        *art : zim.ZimArticle
            Pointer to a C++ article object
        """

        if self.c_read_article != NULL:
            del self.c_read_article

        # Set new internal C zim.ZimArticle article
        self.c_read_article = art

    @staticmethod
    cdef from_read_article(Article art):
        """Creates a python ZimReadArticle from a C++ zim Article article.
        
        Parameters
        ----------
        art : Article
            A C++ zim Article read with File

        Return
        ------
        article: ZimReadArticle
            A initialiazed ZimReadArticle
        """
        cdef ZimReadArticle article = ZimReadArticle()
        article.__setup(<Article *> &art)
        return article

    def is_redirect(self):
       # call is_redirect
        pass


    @property
    def can_write(self):
        return True

    def get_url(self):
        pass

    def get_title(self):
        return self.c_read_article.getTitle().decode('UTF-8')
    
    def get_mime_type(self):
        return  self.c_read_article.getMimeType().decode('UTF-8')
    
    def get_filename(self):
        return ""
  

    def get_data(self):
        return self.c_read_article.getTitle()
        c_blob = self.c_read_article.getData()
        return  ZimBlob.from_c_blob(&c_blob)
```

> In fact, it seems that there is nothing related to reading a zim File even if the `examples.py` use a `pyzim.ZimReader`.

Unpushed WIP.  The goal is to be able to use the same API regardless of the article type.

> 
> If we assume that this branch contains only things related to writing a zim :
> 
> * Please move all this code in a submodule (`write`/`writer`/`writing`/...).
> * Rename `Article`, `ZimArticle` and `ZimArticleWrapper` to something including the fact they are used to write zim file. (If may not be necessary if you move them in a specific submodule)
> 
> Let's be sure that we understand well on what is the separation between `zim::*` and `zim::writer::*` :
> They are two different libraries. They are in the same repository and compiled in the same `.so` but they have almost nothing in common.
> They could be separated in two different repositories without a lot of change (I think that they share the blob structure and one internal small header class only). This is not plan at all, but you should keep this in mind. Especially this is almost two different bindings to create.

This is understood, but the idea I had at first was a pythonic api that doesn't need to know internal libzim details. If you agree we can totally separate reading from writing.

This will however restrict the property uses, for example

python_content = "hola hola"
read_article.content = python_content

I will have to replace the content, currently obtained from the article, with a python constructed Blob. Any ideas on this?

pyzim/pyzim.pyx
jdcaballerov: Can be moved. ✅

pyzim/pyzim.pyx
jdcaballerov: Agree. It's currently there on the WIP.

pyzim/pyzim.pyx
jdcaballerov: I think in this case we can do it.

pyzim/pyzim.pyx
jdcaballerov: Mistake.
```python3
try:
            self.c_creator.addArticle(art)
        except:
            raise
        else:
            if not article.is_redirect:
                self._update_article_counter(article)
```

pyzim/pyzim.pyx
jdcaballerov: error.

pyzim/pyzim.pyx
```diff
@@ -1,338 +1,183 @@
...
-
+        return self.c_article.getMimeType().decode('UTF-8')
+    
     @property
-    def is_redirect(self):
-        """Get if the article is a redirect"""
-        return self.c_zim_article.isRedirect()
+    def content(self):
+        blob = self.c_article.getData()
+        return blob.data()[:blob.size()].decode('UTF-8')
```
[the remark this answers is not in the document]
jdcaballerov: The WIP works different.

pyzim/pyzim.pyx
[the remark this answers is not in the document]
mgautierfr: I'm not sure of what you call the "blend" approach.
If you means that there is only one python article to wrap the two cpp articles, I disagree with this approach.

> The goal is to be able to use the same API regardless of the article type.

That is the main point. We need different API because it is a different use case.

> If you agree we can totally separate reading from writing

I totally agree, this is what I want. This should be handled differently and separately.


> ```
> python_content = "hola hola"
> read_article.content = python_content
> ```

Either I don't understand your question or you don't understand the use case :)
A "read" article (zim::Article) is read only because it is a "view" of what is inside the zim file. You cannot change it's value. So you cannot change it's value using property (or anything else).
On the opposite side a "write" article is not a article but an interface the user code have to implement. So, no property neither.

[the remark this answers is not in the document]
pirate: I'm curious to hear more about your reasoning for keeping the reader and writer separate in python @mgautierfr. To me it seems like most python users would intuitively try to use the same interface, e.g. a typical housekeeping process might be rewriting a few broken image URLs in a ZIM collection:

```python
from libzim import ZimArticle

...
def rewrite_img_urls(zim: ZimArticle):
    article_body = zim.content().decode('UTF-8')
    new_article_body = re.replace(article_body, ... some regex pattern ...)
    zim.content = new_article_body

for article in load_articles(...):
    rewrite_img_urls(article)
```

mgautierfr:
(I move the discussion about the global architecture here, it is better than on a comment on a line).

> I'm curious to hear more about your reasoning for keeping the reader and writer separate in python @mgautierfr. I can understand the reasoning for keeping them separate in C++,

For the same reasoning for keeping them separated in C++. They are two different things not related in anyway (except having almost the same name and method names.) One is a read only object mapping binaries data. The other one is a interface to implement.
You can implement the interface using the read only object but this is a non sense to make it in the wrappper (who is a library from the python user code point of view)
Please, yet again, think as if they were two different libraries in two different repositories.

> but to me it seems like most python users would intuitively try to use the same interface, e.g. a typical housekeeping process might be rewriting a few broken image URLs in a ZIM collection

No, most of the user will read only a zim file (kiwix-desktop, kiwix-serve, kiwix-android, apple/ios application, zim-dump, ...) or write a zim file (zimwriterfs, mwoffline, sotoki,...) without using a existing zim file.
The only tools that use both are zimpatch, zimdiff and zimrecreate. I honestly not sure we should use zimdiff/zimpatch and zimrecreate is mostly a dev tool used to compare zim files created with two version of the libzim's creator.

>  typical housekeeping process might be rewriting a few broken image URLs in a ZIM collection:

This use case is not clear to me. Are you speaking about a tools that change/update zim files to change images url in the articles ?
If yes, this is wrong : You cannot change the content of a zim file. At best you can create a new zim file with different content than the original, but you can't change the zim file itself.

jdcaballerov:
> (I move the discussion about the global architecture here, it is better than on a comment on a line).
> 
> > I'm curious to hear more about your reasoning for keeping the reader and writer separate in python @mgautierfr. I can understand the reasoning for keeping them separate in C++,
> 
> For the same reasoning for keeping them separated in C++. They are two different things not related in anyway (except having almost the same name and method names.) One is a read only object mapping binaries data. The other one is a interface to implement.
> You can implement the interface using the read only object but this is a non sense to make it in the wrappper (who is a library from the python user code point of view)
> Please, yet again, think as if they were two different libraries in two different repositories.

So you will never have a case where a user reads content from a zimfile and wants to write it maybe modifying URL from the python wrapper ?

> 
> > but to me it seems like most python users would intuitively try to use the same interface, e.g. a typical housekeeping process might be rewriting a few broken image URLs in a ZIM collection
> 
> No, most of the user will read only a zim file (kiwix-desktop, kiwix-serve, kiwix-android, apple/ios application, zim-dump, ...) or write a zim file (zimwriterfs, mwoffline, sotoki,...) without using a existing zim file.

If this is the case we can completely separate. OR only.

> The only tools that use both are zimpatch, zimdiff and zimrecreate. I honestly not sure we should use zimdiff/zimpatch and zimrecreate is mostly a dev tool used to compare zim files created with two version of the libzim's creator.
> 
> > typical housekeeping process might be rewriting a few broken image URLs in a ZIM collection:
> 
> This use case is not clear to me. Are you speaking about a tools that change/update zim files to change images url in the articles ?
> If yes, this is wrong : You cannot change the content of a zim file. At best you can create a new zim file with different content than the original, but you can't change the zim file itself.

mgautierfr:
> So you will never have a case where a user reads content from a zimfile and wants to write it maybe modifying URL from the python wrapper ?

zim file are read only. So you will never have a case where a user reads content from a zimfile and write it modified. From python, c++, node or whatever.

One use case, and not the common one, is a user create a new zimfile using the content of another zimfile.
In this case, and only this one, the user code may implement the "writer article" (the interface) with a custom article that use a "read article" to get the data. But the "link" between the writer article and read article is entirely in the user code and never in the wrapper.
You can have a look on zimrecreate (https://github.com/openzim/zim-tools/blob/master/src/zimrecreate.cpp#L34). The article is a `zim::writer::Article` and contains a `zim::Article`. But the link is done in the user code. It is not done in the library (and must not be done there.)

> If this is the case we can completely separate. OR only.

Yes, we must separate them.
It is a OR only for your point of view (library creator). If the user want to reuse a read article as source of writer article, this is it problem and you don't have to know about it.

> Read an article form a zim file, pass it as it is to a zim creator. (no modifications, of any kind)

This is a usecase out of the scope of the wrapper. If a user want to do this (with or without modification) it is his problem, not your.


--- 

This is the same as if you where creating a http library. This library allow the user code to create a client (do requests and receive responses) and also a server (handle requests and create responses). A user may want to create a proxy (handling incoming request by doing outgoing request and using the response it receives to create a responses to send). In this case, the proxy create the outgoing request using the data from the incoming request and it creates the response to send using the data from the response it has received. But all this is made on user code side. The library doesn't know about this and doesn't have to do something specific for this usecase.

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+        content =  blob.data()[:blob.size()]
+        try:
+            return content.decode('UTF-8')
+        except UnicodeDecodeError:
+            return content
+
+    # This changes with implementation
+    @property
+    def can_write(self):
+        raise NotImplementedError
```
mgautierfr, line 88: This should be deleted. Probably a relic of the previous implementation.

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+
+cdef class ZimBlob:
+    cdef Blob* c_blob
+
+    def __init__(self, content):
+
+        if isinstance(content, str):
+            ref_content = content.encode('UTF-8')
+        else:
+            ref_content = content
```
mgautierfr, line 26: We should not assume anything about the content.
It is bytes and we don't want to try to encode a string.
(Explicit is better than implicit)

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+
+        self.c_blob = new Blob(<char *> ref_content, len(ref_content))
```
mgautierfr, line 28: We must keep a reference to `ref_content`.
Else, it may be deleted by the gc and the blob will point to nothing.

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+            raise 
+        else:
+            error[0] = 0
+            value = func()
+            return value.encode('UTF-8')
+    
+    Blob blob_cy_call_fct(void *ptr, string method, int *error):
+        """Lookup and execute a pure virtual method on ZimArticle returning a Blob"""
+        cdef ZimArticle art = <ZimArticle>(ptr)
+        cdef ZimBlob blob = ZimBlob(b'\x00')
```
mgautierfr, line 110: Please do not create a ZimBlob with a dummy bytes.
It is not necessary (it ends by allocating a new `zim::Blob`) and it lead the reader to wrongly assume the bytes is useful.

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+        """Lookup and execute a pure virtual method on ZimArticle returning an int"""
+
+        cdef ZimArticle art = <ZimArticle>(ptr)
+        try:
+            func = getattr(art, method.decode('UTF-8'))
+        except AttributeError:
+            error[0] = 1
+            raise 
+        else:
+            error[0] = 0
```
mgautierfr, line 144: Maybe move all this duplicate code in another function and specialize only the actual call and the type conversion.

libzim/lib.cxx
```diff
@@ -0,0 +1,242 @@
...
+
+void ZimCreatorWrapper::addArticle(std::shared_ptr<ZimArticleWrapper> article)
+{
+    _creator->addArticle(article);
+}
+
+void ZimCreatorWrapper::finalize()
+{
+    _creator->finishZimCreation();
+    delete this;
```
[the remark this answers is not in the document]
jdcaballerov: That slipped into the design from reusing the architecture of node-libzim. The same is done.

https://github.com/openzim/node-libzim/blob/1dbb42c058ed93b3d135aea5e08fec851f0149c6/src/ZimCreator.cc#L61

libzim/libzim.pyx
jdcaballerov: I am using can_write to safeguard writing incomplete articles in the creator. 

Each type of article has different demands as I understand. For instance, redirect can have empty content but needs a redirect url. If someone inherits from this and doesn't implement can_write it will blow before getting to add_article that can cause issues once inside libzim.

libzim/libzim.pyx
jdcaballerov: This allows to construct the blob either from bytes or str. So that when implementing articles you can do:

```python3
 def get_data(self):
        return ZimBlob("<html> hola <\html>")
```
and will work, otherwise you will need to do for normal content strings as most html is.

 def get_data(self):
        return ZimBlob("<html> hola <\html>".encode('UTF-8'))

libzim/libzim.pyx
jdcaballerov: As far as I know when I pass to a parameter to a cython function it auto creates the reference when I do assignment
Please take a look here:
https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#python-objects-as-parameters-and-return-values

that's the reason I don't type it.

I also assign to ref_content so that this protects

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#caveats-when-using-a-python-string-in-a-c-context

libzim/libzim.pyx
jdcaballerov: Agree with you, but I can't declare a cdef inside conditionals (cython limitation) it should be at  the start, I need to have an "empty object"  to capture the returned Blob.  Maybe you can help me figure out a less elegant way to do it while respecting Cython's requirement.

```python3
Blob blob_cy_call_fct(void *ptr, string method, int *error):
        """Lookup and execute a pure virtual method on ZimArticle returning a Blob"""
        cdef ZimArticle art = <ZimArticle>(ptr)
        cdef ZimBlob blob = ZimBlob(b'\x00')
        try:
            func = getattr(art, method.decode('UTF-8'))
        except AttributeError:
            error[0] = 1
            raise 
        else:
            error[0] = 0
            blob = func()
            return dereference(blob.c_blob) 
```

libzim/lib.cxx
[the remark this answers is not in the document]
jdcaballerov: Fix https://github.com/openzim/python-libzim/pull/3/commits/867a170d308a6e6b9ed5adabda1ecc123fb9934b

libzim/lib.cxx
```diff
@@ -0,0 +1,242 @@
...
+
+    virtual zim::writer::Url getMainUrl()
+    {
+        return zim::writer::Url('A', mainPage);
+    }
+
+    void setMainUrl(std::string newUrl)
+    {
+        mainPage = newUrl;
+    }
```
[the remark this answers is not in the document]
jdcaballerov: I'll create an issue so we address it after. https://github.com/openzim/python-libzim/issues/11

libzim/libzim.pyx
```diff
@@ -0,0 +1,286 @@
...
+    @property
+    def main_page(self):
+        """Get the main page of the ZimCreator object"""
+        return self.c_creator.getMainUrl().getLongUrl().decode("UTF-8", "strict")[2:]
+    
+    @main_page.setter
+    def main_page(self,new_url):
+        """Set the main page of the ZimCreator object"""
+        # Check if url longformat is used
+        if new_url.find('/') == 1:
```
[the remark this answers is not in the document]
jdcaballerov: agree. fix https://github.com/openzim/python-libzim/pull/3/commits/e1e4ac758dc6dbb171ff11739d3ae5c1a81f6ae8

libzim/libzim.pyx
jdcaballerov: Fix https://github.com/openzim/python-libzim/pull/3/commits/a8c5bafd4f5ea761190844d1097ba73d55033c5f

libzim/libzim.pyx
jdcaballerov: Done. https://github.com/openzim/python-libzim/pull/3/commits/cc31d318f78cc6eccf3091d692d7d4330eb60988

libzim/libzim.pyx
mgautierfr: Hum. Redirect articles must return true for `isRedirect`. If this is the case, libzim will only use the `getRedirectUrl` and not `getData`.

I'm not sure of what you want to protect with `can_write`. Is `can_write` is equivalent to `isRedirect` ? Or you want to protect of something else.

libzim/libzim.pyx
mgautierfr: Hum, yes.
I tend to have all the API "exchanging data" to use bytes. Because this is how things are exchanged and you cannot assume (most of the time) the encoding.
However, the zim format specify that the encoding must be utf8, so we are somehow on a specific case here.
Let's move on with this, we keep the implicit encoding of a str.

rgaudin: Well this implicit dual case forces the user to also handle the two cases. Could we keep this api call bytes only and maybe add a `text_content` that automatically decodes UTF-8?

libzim/libzim.pyx
mgautierfr: The "auto reference increment" of cython works well when your function only use the parameter:
- increment ref to pyobject
- "wrap" pyobject into a cobject
- pass cobject to function
- use cobject in function
- quit function
- decrement ref to pyobject.

But now we are not directly using it. We are storing the cobject to use it later:
- increment ref to (ref_content) pyobject
- "wrap" pyobject into a cobject
- pass cobject to (c)Blob constructor
- store cobject in (c)Blob
- quit constructor
- decrement ref to (ref_content) pyobject.
- Store the (c)blob in (py)Blob
- quit (py)Blob.__init__ (and so left all reference to `ref_content`)
- [Later] Use Blob. <--- But ref_content doesn't exist.

I think we have to keep the `ref_content` in (py)Blob should be enough (so ref_content exist as long as (py)Blob exist).
And we sure that we keep a reference to (py)Blob when we pass the (c)Blob to the c world.
The best is probably to use a intermediate method in `Article` (it takes the (py)Blob return by the `get_data` and keep a ref in the article before returning it (where the blob will be converted to c in `blob_cy_call_fct`).
It will be good as long as libzim doesn't try to read the blob after the (py)Article has been delete.
Libzim ensure that it will keep a reference to the std::shared_ptr<zim::creator::Article>` as long as it need it (and so use blob).
So the wrapper should be sure that we keep a reference to the (py)Article as long as the `zim::creator::article` is not deleted. (But I think we are good here because the Article wrapper increment the ref on the python article).

libzim/libzim.pyx
mgautierfr: We are on the writer part here (the `get_data` is called by the writer to know the data to write in the zim file).

But yes, on the reader part, we shouldn't have a method that some time return bytes and sometime str.
The "base" method should return only bytes (without trying to decode anything) and we could add a "convenient" method to get a decoded str.
But I'm not sure we need to have a second method. It will be explicit in any case (call the base method or the convenient one) and if it is explicit, it is better to let the user code call `decode()` when it needs it.
Anyway, this is a discussion to have one the reader part, not here :)

libzim/libzim.pyx
jdcaballerov: I want to safeguard ZimCreator.add_article. From my point of view it's not 100% equivalent. 
For instance I was requiring a normal html article to have a Url and content. Redirect articles can required to have redirect url but can have empty content.

mgautierfr: Yes, but I don't understand when `can_write` would be false or true. The user code will always return true no ?
Having a `can_write` doesn't protect you from articles badly implemented.

mgautierfr:
This PR is too long and too big. Especially if you add the reading part at the end.
I have :
- Rewrite a bit the git history to remove unwanted commit (small fix of previous commits) and push the changes into master (only the writing part).
- Create a new PR (https://github.com/openzim/python-libzim/pull/16) for the reading part.
- Open issues on things that have to be fixed, even if the code is in master. (#12, #13, #14, #15, #17). #17 is a  bonus (that have to be done) but the other are real bug and must be fixed before using the writer in sotoki (or elsewhere)

Issue #6, opened 2020-04-09
Add getMainPage and setter the Creator wrapper


jdcaballerov:
> Probably upstream limitation but requiring main_page on ZimCreator creation seems weird. You allow to set it to blank but can't change it afterwards?

I used the same construct as node-libzim that fixes it from creation.
ZimCreator must implement a function getMainPage so that when finishZimCreation() (C++) is called the header is filled with it fillHeader().

> With the current design I can overload the wrapper function and add a setter if needed.

I think it makes sense, yes.

jdcaballerov:
Commit https://github.com/openzim/python-libzim/commit/9e45d5c9b905a79768942012998ed6d4ccc7cbe6 Adds the feature

PR #10, opened 2020-04-17
Add Github Actions CI/CD for build, tests, and PyPI release

pirate:
### Key Points

- does everything in Github Actions `ubuntu` environment, does not use standalone Dockerfile (to match style use by other OpenZIM projects)
- installs C++ and Cython build environment, and builds `python-libzim` against `libzim` binary release dylib
- checks code style and syntax with `black` and `flake8`
- runs `python-libzim` [unit test suite](https://github.com/openzim/python-libzim/pull/3/files#diff-9c722b9e26d746524b3a840722f17659) with `pytest` (also supports `tox`)
- automatically pushes any new tagged releases to PyPI (using pypi API key)

### TODOs

- [X] setup PyPI API key in repository secrets for `pypa/gh-action-pypi-publish` to use
- [X] limited support to Python 3.6, 3.7, and 3.8
- [x] remove `libzim` compilation in favor using the latest released binary
- [x] setup `pyproject.toml` + `pytest` + `tox` for Python tests, cython build, sdist/bdist packaging, and PyPI release process
  - https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
  - https://realpython.com/pypi-publish-python-package/
  - https://tox.readthedocs.io/en/latest/example/basic.html#pyproject-toml-tox-legacy-ini
  - https://setuptools.readthedocs.io/en/latest/setuptools.html
  - https://www.bernat.tech/pep-517-518/
  - https://github.com/FedericoStra/cython-package-example
  - https://stackoverflow.com/questions/6633624/how-to-specify-header-files-in-setup-py-script-for-python-extension-module
  - https://cython.readthedocs.io/en/latest/src/tutorial/clibraries.html#static-linking
  - https://cython.readthedocs.io/en/latest/src/tutorial/clibraries.html#dynamic-linking
- [x] ~~set up statically linked cross-platform distribution with some combination of `cibuildwheel`, `manylinux`, `auditwheel`~~ (not needed for now, just do binary-based Linux x64 build cc @kelson)
  - add libzim build process back to github action? (needed to cross-platform compilation)
  - https://github.com/pypa/auditwheel
  - https://github.com/grantjenks/python-c2f
  - https://github.com/pypa/manylinux
  - https://github.com/RalfG/python-wheels-manylinux-build
  - https://github.com/FedericoStra/cython-package-example


rgaudin:
* Please remove py35, we don't want to support it.
* Didn't @kelson42 asked to build using the released libzim binary?

pirate:
We don't mind changing it to using the pre-built release of `libzim`, but then we wont be able to test with CI/CD on macOS/Windows automatically, as there are only releases available for Linux at the moment here: http://download.openzim.org/release/libzim/libzim_linux-x86_64-6.1.1.tar.gz

afreydev:
Cython has some problems with in versions > 0.29.6
https://github.com/pandas-dev/pandas/issues/26127

MANIFEST.ini
```diff
@@ -0,0 +1,3 @@
+include LICENSE
+include README.md
+recursive-include libzim_linux-x86_64-6.1.1 *
```
pirate, line 3: This lets us optionally ship the libzim.so binary with the bindings it's compiled against it, so the user doesn't have to install it separately, and doesn't risk accidentally installing a different version system-wide than the ones the bindings were compiled against. (it will still use the system-installed dylib first, but the user can force it to use this inbuilt shipped one with some env vars if they don't want to install it separately)

Let me know if this is ok, or if you want us to remove this and force the user to always install libzim on the system separately.

mgautierfr:
Please rebase this PR on master, not on cython-blend.

pirate:
@mgautierfr this is rebased on `master` and ready for review.  The tests are failing until `libzim_api.h` is merged into master from #3, but almost everything else should work as expected.

I suggest we try to get this merged into master pretty much as-is (small suggestions welcome of course), and work on any big changes as separate PRs against master. It will make it easier to coordinate work in other branches since this PR changes the testing system a little and the other branches need that work to continue.

.github/workflows/test.yml
```diff
@@ -0,0 +1,85 @@
...
+      
+      - name: Sync libzim dylib & headers in system
+        run: |
+          sudo rsync -a $GITHUB_WORKSPACE/libzim_linux/$LIBZIM_LIBRARY_PATH /usr/lib/libzim.so
+          sudo rsync -a $GITHUB_WORKSPACE/libzim_linux/$LIBZIM_INCLUDE_PATH/ /usr/include/zim/
+      
+      - name: Build cython, sdist, and bdist_wheel
+        run: |
+          pip install --upgrade cython==$CYTHON_VERSION setuptools pip wheel
+          python3 setup.py build_ext --build-lib=cython_build
```
pirate, line 76:
```suggestion
          python3 setup.py build_ext
```

mgautierfr:
> @mgautierfr this is rebased on master and ready for review. The tests are failing until libzim_api.h is merged into master from #3, but almost everything else should work as expected.

`libzim_api.h` is generated by Cython. It should not be in the repository. In fact, I have remove them from the code in master and in the PR #6.

.github/workflows/release.yml
```diff
@@ -0,0 +1,194 @@
...
+        if: steps.cache-libzim.outputs.cache-hit != 'true'
+        run: |
+          wget -q https://download.openzim.org/release/libzim/$LIBZIM_RELEASE.tar.gz
+          tar --gunzip --extract --file=$LIBZIM_RELEASE.tar.gz
+          mv $LIBZIM_RELEASE libzim_linux
+
+      - name: Sync libzim dylib & headers in system
+        run: |
+          sudo rsync -a $GITHUB_WORKSPACE/libzim_linux/$LIBZIM_LIBRARY_PATH /usr/lib/libzim.so
+          sudo rsync -a $GITHUB_WORKSPACE/libzim_linux/$LIBZIM_INCLUDE_PATH/ /usr/include/zim/
```
mgautierfr, line 58: It would be easier to set the CPPFLAS to `-I$GITHUB_WORKSPACE/libzim_linux/$LIBZIM_INCLUDE_PATH -L$GITHUB_WORKSPACE/libzim_linux/$LIBZIM_LIBRARY_PATH` instead of copying file in the (root) fs

.github/workflows/release.yml
```diff
@@ -0,0 +1,194 @@
...
+          python3 setup.py sdist bdist_wheel
+          python -m cibuildwheel --output-dir wheelhouse
+
+      - uses: actions/upload-artifact@v1
+        with:
+          name: wheels
+          path: ./wheelhouse
+
+      - name: Push release to PyPI
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
mgautierfr, line 77: It seems that it will publish on PyPI on all push on master. Not only when we tag a release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,194 @@
...
+  #         cp macOS-wheels/*.whl dist/
+  #         cp Windows-wheels/*.whl dist/
+  #         ls -la dist/
+  #         sha256sum dist/*.whl
+
+  #     - uses: pypa/gh-action-pypi-publish@master
+  #       if: github.event_name == 'release'
+  #       with:
+  #         user: __token__
+  #         password: ${{ secrets.pypi_token }}
```
mgautierfr, line 194: Please use another branch if you have a alternative approach.
And git allows us to do a diff between the to approaches.

.github/workflows/test.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.6
+          architecture: x64
+
+      - name: Autoformat with black
+        run: |
+          pip install black
+          black . --exclude=setup.py
```
mgautierfr, line 26: This should fails if black reformat files.
If you just reformat the files on the CI it is useless.

.github/workflows/test.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Lint with flake8
+        run: |
+          pip install flake8
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --exclude=setup.py --show-source --statistics
+          # exit-zero treats all errors as warnings, this can be made stricter later
+          flake8 . --count --exit-zero --exclude=setup.py --max-line-length=$MAX_LINE_LENGTH --statistics
```
mgautierfr, line 34: Same here. We want to CI to pass only if everything is correct.

MANIFEST.ini
mgautierfr: No it is good to include the libzim.so binary in the bdist.

However, I don't understand the `*` at the end of the line. We should include only the `.so` no ?

setup.py
```diff
@@ -1,30 +1,168 @@
...
+GITHUB_URL = "https://github.com/openzim/python-libzim"
+
+BASE_DIR = Path(__file__).parent
+LIBZIM_CYTHON_DIR = 'libzim'
+
+LIBZIM_INCLUDE_DIR = os.getenv('LIBZIM_INCLUDE_DIR')
+LIBZIM_LIBRARY_DIR = os.getenv('LIBZIM_LIBRARY_DIR')
+
+INCLUDE_DIRS = [str(BASE_DIR / LIBZIM_CYTHON_DIR), '/usr/include']
+LIBRARY_DIRS = ['/usr/lib']
```
mgautierfr, line 54: You should not assume the default include and library dir.
Depending on the system, the default lib dir can be `/usr/lib`, `/usr/lib64`, `/usr/lib/linux-x86_64`, ...

setup.py
```diff
@@ -1,30 +1,168 @@
...
 
-# Utility function to read the README file.
-# Used for the long_description.  It's nice, because now 1) we have a top levegit checkout masterl
-# README file and 2) it's easier to type in the README file than to put a raw
-# string in below ...
-def read(fname):
-    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+
+PACKAGE_NAME = "libzim"
+VERSION = "6.1.1"  # pegged to be the same version as libzim since they are always released together
```
mgautierfr, line 40: THis shouldn't be the same.
A minor version change of libzim doesn't need a new version of the wrapper.
And you may want to fix things in the wrapper even if libzim doesn't change.

tests/test_libzim.py
```diff
@@ -74,45 +58,83 @@ def get_data(self):
...
-        rnd_str = str(uuid.uuid1()) 
-        zim_creator = ZimCreator(self.test_zim_file_path + '-' + rnd_str + '.zim',main_page = "welcome",index_language= "eng", min_chunk_size= 2048)
+        zim_creator = ZimCreator(
+            os.path.join(self.test_dir, f'test-metadata-{uuid.uuid1()}.zim'),
+            main_page="welcome",
+            index_language="eng",
+            min_chunk_size= 2048,
+        )
         zim_creator.update_metadata(**TEST_METADATA)
         self.assertEqual(zim_creator._get_metadata(), TEST_METADATA)
```
mgautierfr, line 93: Don't forget to call finalize.

Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+pytest = "*"
+tox = "*"
```
mgautierfr, line 8: Why are we using tox ?
I don't really know it but it seems its doing the same as pipenv.

.github/workflows/release.yml
pirate: This section at the top combined with this section should limit PyPI releases to only push tags on master starting with `v*`:
```python3
on:
  release:
    types: [published]
    tags:
      - v*
```

.github/workflows/release.yml
pirate: This is unlikely to be merged for many months, so keeping it as a separate branch would accumulate merge conflicts over time. Can I put it in the Github Wiki or a separate document instead maybe?

.github/workflows/test.yml
pirate: yup, I know, see the `this can be made stricter later` comment below. we're going to upgrade it to fail on reformatting in a later commit, we just wanted to make sure the rest of the CI process can pass successfully, we don't intend to keep it this way

MANIFEST.ini
pirate: The headers must be included in the `sdist` (because they are needed for Cython to be able to run), but they are not included in the `bdist` (which does not need the headers because it's already compiled into a self-contained .so).

Pipfile
pirate: It was useful for testing across multiple Python versions automatically (which `pipenv` cannot not do on its own), but I took it out because we don't *really* need it.

.github/workflows/release.yml
pirate: We have to `mv` it anyway because `cython` can't find the `libzim.so` file via the `libzim.so.6.1.1` symlink for some reason, it only works when it's named `libzim.so`.

Installing it system-wide also makes it more convenient later as python finds it automatically during the `setup.py` Cython and `cythonize` process.

setup.py
pirate: In the NPM world it's common to pin the binding package version to the same version as the main library, and then use a script to bump the versions of both whenever either one wants to issue a new release. (e.g. in this way `redux==1.2.3` is always guaranteed to work with `react-redux==1.2.3`,  `react-redux-addons==1.2.3`, etc. and users don't have to look up compatibility charts or wonder if they have the correct versions installed)

I don't have a strong preference either way though, so I don't mind changing it.

setup.py
pirate: changed it to use a relative `./lib` and `./include`

tests/test_libzim.py
pirate, line 93:
```suggestion
        self.assertEqual(zim_creator._get_metadata(), TEST_METADATA)
        zim_creator.finalize()
```

.github/workflows/release.yml
pirate: this is now fixed to use relataive `./lib` and `./include` in the github workspace dir

.github/workflows/test.yml
pirate: these will now fail on format errors

pirate:
Note: the build is failing because of these two files.

```
would reformat /home/runner/work/python-libzim/python-libzim/examples/basic.py
would reformat /home/runner/work/python-libzim/python-libzim/tests/test_libzim.py
```
I don't want to reformat these files with black in this PR, as it will create a lot of unnecessary conflicts with our changes to those files in other branches.
We should wait till there is a complete stable version merged in master, then reformat all the code at once in one PR.

jdcaballerov:
> > @mgautierfr this is rebased on master and ready for review. The tests are failing until libzim_api.h is merged into master from #3, but almost everything else should work as expected.
> 
> `libzim_api.h` is generated by Cython. It should not be in the repository. In fact, I have remove them from the code in master and in the PR #6.

It is [recommended to include the generated files](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules) so that the user doesn't need to have Cython installed.

kelson42:
> > > @mgautierfr this is rebased on master and ready for review. The tests are failing until libzim_api.h is merged into master from #3, but almost everything else should work as expected.
> > 
> > 
> > `libzim_api.h` is generated by Cython. It should not be in the repository. In fact, I have remove them from the code in master and in the PR #6.
> 
> It is [recommended to include the generated files](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules) so that the user doesn't need to have Cython installed.

Nothing which is automatically generated should be in the git. Cython is a strong dependence and it should be clearly written in the `README.md`, but any developer has to have it installed.

MANIFEST.ini
mgautierfr: I would say that we don't want to include the library (and its header) in the sdist.
If the user want the source, it is to develop and he can compile the library itself (or dl a official release).
Including the `.so` in the bdist is useful to just `pip install` and use.

The common scheme is to not include dependencies in the sources (whatever the project language is).

.github/workflows/release.yml
mgautierfr: > because cython can't find the libzim.so file via the libzim.so.6.1.1 symlink for some reason, it only works when it's named libzim.so

It would be good why. "Packaged installation" doesn't provide a libzim.so but only a versioned library (`so.6.1.1`, `so.6`, ...).
The plain unversionned `.so` is used a link time. (It is why it is in the `-dev` packages).

.github/workflows/release.yml
mgautierfr: Yes, but in the same time you write just after :
```
# Uncomment this if you'd like to test the build+release process on PRs too
  # (it will only test the build, it wont push to PyPI until a new release is tagged)
  # push:
  #   branches:
  #     - master
  # pull_request:
  #   branches:
  #     - master
```
If I uncomment it, ci will push the "non release" on PyPI.

.github/workflows/release.yml
mgautierfr: It will be a branch with change only on `.github/workflows/release.yml`. There is few chance to have conflict. And if it there are, having the code commented in the same file will not help.

setup.py
mgautierfr: How do they do in NPM to update the wrapper even if the library has not changed ?

Pipfile
mgautierfr: Yes, especially if the CI matrix does the same.

setup.py
```diff
@@ -1,30 +1,151 @@
...
+# the default is to dynamically link (finds externally installed libzim on user's system)
+if not (LIBZIM_LIBRARY_DIR or find_library('zim')):
+    print(
+        '[!] Warning: Could not find libzim.so in available system libraries or LIBZIM_LIBRARY_DIR\n'
+        '    Hint: Install it from source from https://github.com/openzim/libzim\n'
+        '          or download a prebuilt zim release and set the env varaibles to point to it:\n'
+        f'    LIBZIM_INCLUDE_DIR=/libzim_linux-x86_64-{VERSION}/include\n'
+        f'    LIBZIM_LIBRARY_DIR=/libzim_linux-x86_64-{VERSION}/lib/x86_64-linux-gnu'
+    )
+
```
mgautierfr, line 76: All this thing about the diretories to find the libzim seems a bit to complex.
The user can simply set `CPPFLAGS=-I/.../include -L/.../lib` before running `setup.py` and everything is working fine.

To launch the wrapper, the user has to set `LD_LIBRARY_FLAGS=.../lib:$LD_LIBRARY_FLAGS`.

And so, there is nothing to do on setup.py to handle where the library is.

setup.py
pirate: They have a script to bump both versions anytime either one changes.

setup.py
pirate: `CPPFLAGS` is pretty foreign to the average python user, I was trying to get it so that even `LIBZIM_LIBRARY_DIR` doesn't need to be set most of the time.

In the interest of time though I can remove some of this autodetection in favor of requiring the user use more env vars.

pirate:
> Nothing which is automatically generated should be in the git. Cython is a strong dependence and it should be clearly written in the README.md, but any developer has to have it installed.

@kelson42 this is the case right now because we're only distributing `bdist` (binary builds), but assuming the goal is to change that in the future, the recommended best practice is to ship the headers and autogenerated cython code in the `sdist` so that users who want to install from source can do so without needing a whole Cython developer environment.

If you are 100% you don't want users installing via `sdist` and force PyPI users to *only* use the `bdist` binary distribution, then we can remove it, but note that this is fairly unusual for PyPI projects (where the assumption is that all projects have both an `sdist` and `bdist`).

.github/workflows/release.yml
pirate: I don't know any more detail unfortunately, other than the fact that if you set LD_LIBRARY_PATH to a folder that only contains the `.so.6.1.1` and `so.6` it doesn't work, but if it contains the `.so` then it works.

.github/workflows/release.yml
pirate: Yes, that's intentional as it's useful for releases to test PyPI `https://test.pypi.org/legacy/`, where you can push a release on every PR and push in order to make sure the build/release process works correctly.  If you want me to remove it I can, the alternative is to just test locally with test.pypi.org and not use Github Actions for test releases.

pirate: (removed)

.github/workflows/release.yml
pirate: Moved to: https://github.com/openzim/python-libzim/issues/25

rgaudin:
> If you are 100% you don't want users installing via `sdist` and force PyPI users to *only* use the `bdist` binary distribution, then we can remove it, but note that this is fairly unusual for PyPI projects (where the assumption is that all projects have both an `sdist` and `bdist`).

sdist is not a source release, it’s a source distribution for setup tools which is not a complex system like debian debs. Of all the dependencies I’ve used that required install-time compilation, none ever required cython or similar.
I think @kelson42 ’s comment was general as he may not be familiar with this. 
As a reminder, this is generated and put into the sdist archive but not stored in the repo obviously.

pirate:
Ok so just to be clear, the conclusion is:

- keep the autogenerated files in the `sdist` so that cython is not needed during `install-time compilation`
- don't keep the autogenerated files in git (they must be regenerated anytime a new `sdist` is rolled, and the machine doing the release will have Cython as a requirement)

I can push the change today.

rgaudin:
> Ok so just to be clear, the conclusion is:
> 
> 
> 
> - keep the autogenerated files in the `sdist` so that cython is not needed during `install-time compilation`
> 
> - don't keep the autogenerated files in git (they must be regenerated anytime a new `sdist` is rolled, and the machine doing the release will have Cython as a requirement)
> 
> 
> 
> I can push the change today.

@mgautierfr makes the decision but that’s how it should be done in my opinion, yes. that’s the expected python/pip behavior.

pirate:
Ok that should now be done √:
**`.gitignore`:**
```bash
...
# Files autogenerated by Cython
libzim/libzim.cpp
libzim/libzim.h
libzim/libzim_api.h
```
**`MANIFEST.in`:**
```bash
...
# include the cython-generated libzim and libzim_api cpp+headers in the sdist
recursive-include libzim *
```

setup.py
pirate: This is done ee61d2a

setup.py
pirate: Want me to reset this to 0.0.1, or keep it as is?

rgaudin: FYI the project name on pypi has been reserved by making a `0.0.1` so whatever version we release first much be different.

afreydev:
I have opened an issue in libzim project related to the segmentation fault problem in the test github actions workflow:
https://github.com/openzim/libzim/issues/333

afreydev:
Would it be possible complete this workflow by using compiled source (instead of pre-built) while this (https://github.com/openzim/libzim/issues/333) is solved? @mgautierfr

mgautierfr:
It would be difficult. The compilation of libzim should not be handle by the wrapper. There is kiwix-build for that.
openzim/libzim#333 should simply be fixed.

---

The PR is pretty old, it should be rebase on master (where the code has changed since). And ideally with the intermediate (wrong) commits removed.

mgautierfr:
Seems good. We may change few things in the future but it is good for now and this PR is open since a long time now.
If I have things to change, I will open another PR.

Issue #14, opened 2020-04-21
finalize must be always called.


mgautierfr:
There are two issues here :
- One on libzim assuming that the finalize method will be call before deletion of the creator.
- `test_libzim.py` not calling the finalize method.

For the second point, the more pythonic way to solve this is to use a contextmanager to force the call to finalize in the `__exit__` method.

PR #18, opened 2020-04-21
Add context manager version of ZimCreator to enforce automatic finalization

pirate:
Usage:

```python3
from libzim import zimcreator, ZimArticle...

with zimcreator(...) as zc:
    zc.add(ZimArticle(...))

# zc.finalize() is called automatically on leaving the context
```

### DIFF
--- a/libzim/libzim.pyx
+++ b/libzim/libzim.pyx
@@ -7,6 +7,8 @@ from libcpp.string cimport string
 from libcpp cimport bool
 from libcpp.memory cimport shared_ptr, make_shared
 
+from contextlib import contextmanager
+
 
 import datetime
 
@@ -345,7 +347,15 @@ cdef class ZimCreator:
     def __repr__(self):
         return f"{self.__class__.__name__}(filename={self.filename})"
 
-
+    
+@contextmanager
+def zimcreator(*args, **kwds):
+    """Context manager version of ZimCreator that calls .finalize() automatically on exit"""
+    zimcreator = ZimCreator(*args, **kwds)
+    try:
+        yield zimcreator
+    finally:
+        zimcreator.finalize()
 
 #########################
 #    ZimFileArticle     # 


kelson42:
@pirate I have added  #14 as linked issue, I guess this is the issue which is fixed here. Thank you very much for finally have made the proper fix in place of just fixing the test. Really much appreciated.

rgaudin:
@kelson I'm not sure this issue will be *fixed* by this PR. This is an additional convenient way of using the creator that calls `finalize()` for you but:

* It's not yet used in `test_libzim.py`
* @mgautierfr mentioned a libzim issue in #14 that may need to be tracked: you can't always use a context manager (especially in our current scrapers) so it's user's responsibility to call it to get expected behavior but are there side consequences in not doing so that should be prevented?

mgautierfr:
It is syntaxic sugar, but I prefer to have a classmethod open as contentmanager (move the function as a classmethod `open` in `ZimCreator`):
```
with Zimcreator.open(...) as zc:
    zc.add(ZimArticle(...))
```
or directly with to constructor (implement `__enter__` and `__exit__` in `ZimCreator` :
```
with Zimcreator(....) as zc:
    ...
```

> you can't always use a context manager (especially in our current scrapers) 

I'm curious to see why.

> so it's user's responsibility to call it to get expected behavior

Yes, if you don't use a context manager you have to close/finalize the object yourself.
It is like for a plain file object open to write in it. If you don't close the file and quit your program, you may have data not written to the fs. 

> but are there side consequences in not doing so that should be prevented?

For now, if you don't call finalize before destroying the creator, you have a sigsev fault in libzim.

mgautierfr:
Please rebase this PR on master.

jdcaballerov:
Suggestions implemented in based of master https://github.com/openzim/python-libzim/pull/24

Issue #12, opened 2020-04-21
ZimBlob must keep a reference on the content.


mgautierfr:
See https://github.com/openzim/python-libzim/pull/3#discussion_r410445830

mgautierfr:
You added a reference to ZimBlob in the article.
But you also need to keep a reference to the content itself in the ZimBlob.

jdcaballerov:
Do you agree ?

```python3
cdef class ZimBlob:
    cdef Blob* c_blob
    cdef object ref_content
    
    def __cinit__(self, content):

        if isinstance(content, str):
            self.ref_content = content.encode('UTF-8')
        else:
            self.ref_content = content

        self.c_blob = new Blob(<char *> self.ref_content, len(self.ref_content))

    def __dealloc__(self):
        if self.c_blob != NULL:
            del self.c_blob
```

mgautierfr:
> Or should I retain only the address of that content. If the content is encoded I guess a copy will be made.
What need to be done in incrementing the reference on `ref_content` to be sure that gc will not delete it.
"Storing" `ref_content` in self is incrementing the reference. (Without copy, you are making a new reference).

But it is fix with my commit https://github.com/openzim/python-libzim/pull/22/commits/dee124f878a78c6961722f03a5d8593943780c0d of the PR #22 
(The same way of your commit https://github.com/openzim/python-libzim/commit/1133c846c79f45749088123f925958b8f4fb02c3 but not merged because on branch cython-blend)

PR #22, opened 2020-04-22
Fix multi thread

mgautierfr:


mgautierfr:
Also fix #12 and #13.

jdcaballerov:
@mgautierfr I think we also need to acquire the GIL at the helper. I tested with this and works great with 100,000 arts compressed and indexed.

https://github.com/openzim/python-libzim/pull/22/commits/e15448ef2ffae1ef1e888c56fac9a07bdb16caf0

PR #35, opened 2020-06-08
build and test on macOS

rgaudin:
- updated test's libzim version to 6.1.5 to match release
- added a macos-latest OS entry
- added separate macOS steps when required
- using `macosx_10_9_x86-64` as platform name.
There doesn't see to be a linuxmany equivalent.
Not much of a problem now as 10.9 introduced 64b only.
- release to PiPy using twine as the ga used was through docker which is linux-only on ga.
- updated descriptions to inform user about macOS availability

This was tested on test.pypi.org (now deleted)

```py
pip install -i https://test.pypi.org/simple/ libzim==0.0.4.dev4
Looking in indexes: https://test.pypi.org/simple/
Collecting libzim==0.0.4.dev4
  Downloading https://test-files.pythonhosted.org/packages/d2/e3/75b9d6d79e2733f6ab02ba316094a993ee658c89c91f2d1901d31557adfb/libzim-0.0.4.dev4-cp38-cp38-macosx_10_9_x86_64.whl (6.8 MB)
     |████████████████████████████████| 6.8 MB 369 kB/s
Installing collected packages: libzim
Successfully installed libzim-0.0.4.dev4
[tw3] reg@homelet:[~/src/github]: $ python
Python 3.8.0 (v3.8.0:fa919fdf25, Oct 14 2019, 10:23:27)
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import libzim.reader
>>> f = libzim.reader.File("/Users/reg/data/zim/disledansmalangue_fr_2020-05.zim")
>>> f.main_page_url
'A/home.html'
>>>
```

This was tested using manually-uploaded libzim binaries to http://tmp.kiwix.org/lzrlz/. This obviously depends on https://github.com/kiwix/kiwix-build/pull/445 and a new, public release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,16 +6,19 @@ on:
       - v*
 
 env:
-  LIBZIM_RELEASE: libzim_linux-x86_64-6.1.5
+  LIBZIM_VERSION: 6.1.6
   LIBZIM_INCLUDE_PATH: include/zim
   CYTHON_VERSION: 0.29.6
+  TWINE_USERNAME: __token__
+  TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+  # TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
 
 jobs:
   release:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest]
         # TODO: expand this to cross-platform builds (see V2 approach below)
         # os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: [3.6, 3.7, 3.8]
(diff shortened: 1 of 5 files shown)


rgaudin:
rebased and updated release version to `6.1.6`

mgautierfr:
I trust you on this PR (at least I've seen no blocker).

Just a remark about the libzim version. There is a "permalink" available to download libzim without version (see https://github.com/kiwix/kiwix-build/issues/448).
It think we can update this PR to not specify a version and so, we will not need to update the CI when we release a new libzim version.

rgaudin:
OK! permalink is not there yet for macos. Will coordinate with @kelson42 and fix this once it's OK.

kelson42:
> It think we can update this PR to not specify a version and so, we will not need to update the CI when we release a new libzim version.

I would really prefer to tackle a specific version to avoid any surprise and just to be able to clear track which version the binding use. This is how we do node-libzim.

mgautierfr:
That's right yes. Then we can merge this.

PR #44, opened 2020-06-10
Properly dealloc creatorWrapper on close

### LINKED ISSUE #31: segfault on variable reuse
Reusing the same variable for different `Creator` segfaults

```py
In [1]: import libzim.writer

In [2]: c = libzim.writer.Creator("/data/test03.zim", "/A/index.html", "fra", 2048)
/data/test03.zim

In [3]: c = libzim.writer.Creator("/data/test04.zim", "/A/index.html", "fra", 2048)
/data/test04.zim

In [4]: Segmentation fault
```

### LINKED ISSUE #41: Zim with no article leave temp files behind
```py
with libzim.writer.Creator("test_x05.zim", main_page="A/index.html", index_language="fra", min_chunk_size=2048) as zim:
    pass

T:7; A:3; RA:0; CA:3; UA:0; FA:0; IA:0; C:0; CC:0; UC:0; WC:0
T:7; Waiting for workers
T:7; ResolveRedirectIndexes
Resolve redirect
T:7; Set article indexes
set index
T:7; Resolve mimetype
T:7; create title index
T:7; 5 title index created
T:7; 2 clusters created
T:7; write zimfile :
T:7;  write mimetype list
T:7;  write directory entries
T:7;  write url prt list
T:7;  write title index
T:7;  write cluster offset list
T:7;  write header
T:7;  write checksum
T:7; rename tmpfile to final one.
T:7; finish
```

and on the filesystem:
```
.rw-r--r--   24k reg  staff 10 Jun 13:33  test_x05.idx
drwxr-xr-x     - reg  staff 10 Jun 13:33  test_x05.idx.tmp
.rw-r--r--   50k reg  staff 10 Jun 13:33  test_x05.zim
.rw-r--r--   24k reg  staff 10 Jun 13:33  test_x05_title.idx
drwxr-xr-x     - reg  staff 10 Jun 13:33  test_x05_title.idx.tmp
```

zim file is working (albeit without actual article) but the xapian stuff is not removed.

rgaudin:
Fixes #41 by explicitly deleting the creatorWrapper once we've finalized it on the `Creator.close`.

Also fixes #31 by calling `close()` on `Creator` (pure-python) deletion

### DIFF
--- a/libzim/writer.py
+++ b/libzim/writer.py
@@ -150,6 +150,9 @@ class Creator:
     def __exit__(self, *args):
         self.close()
 
+    def __del__(self):
+        self.close()
+
     def add_article(self, article):
         self._creatorWrapper.add_article(article)
         if not article.is_redirect():
--- a/tests/test_libzim.py
+++ b/tests/test_libzim.py
@@ -159,3 +159,28 @@ def test_creator_params(tmpdir):
     assert zim.filename == path
     assert zim.main_page_url == main_page_url
     assert bytes(zim.get_article("/M/Language").content).decode("UTF-8") == index_language
+
+
+def test_segfault_on_realloc(tmpdir):
+    """ assert that we are able to delete an unclosed Creator #31 """
+    creator = Creator(str(tmpdir / "test.zim"), "welcome", "eng", 2048)
+    del creator  # used to segfault
+    assert True
+
+
+def test_noleftbehind_empty(tmpdir):
+    """ assert that ZIM with no articles don't leave files behind #41 """
+    fname = "test_empty.zim"
+    with Creator(
+        str(tmpdir / fname), main_page="welcome", index_language="eng", min_chunk_size=2048,
+    ) as zim_creator:
+        print(zim_creator)
+
+    assert len([p for p in tmpdir.listdir() if p.basename.startswith(fname)]) == 1
+
+
+def test_double_close(tmpdir):
+    creator = Creator(str(tmpdir / "test.zim"), "welcome", "eng", 2048)
+    creator.close()
+    with pytest.raises(RuntimeError):
+        creator.close()


rgaudin:
I am unable to reproduce the failing test, neither on macOS (py38, py36) nor on docker/linux (`python:3.8`). I'm short of explanation for this…

libzim/writer.py
```diff
@@ -183,8 +186,10 @@ def _get_counter_string(self):
         return ";".join(["%s=%s" % (k, v) for (k, v) in self._article_counter.items()])
 
     def close(self):
-        self.write_metadata()
-        self._creatorWrapper.finalize()
+        if hasattr(self, "_creatorWrapper"):
+            self.write_metadata()
+            self._creatorWrapper.finalize()
+            del self._creatorWrapper
```
mgautierfr, line 192: I'm not really confident with this `del`. It means that after a `close` the creator will be in an inconsistent state and a call to another method will lead to use a freed memory.

User should not call a method after a `close`, but we have no guaranty here.
If we delete the creator wrapper, we should at least test it before using it.

Or we can simply move the deletion in the `__del__` method.

tests/test_libzim.py
```diff
@@ -140,3 +140,24 @@ def test_check_mandatory_metadata(tmpdir):
             title="Test Zim",
         )
         assert zim_creator.mandatory_metadata_ok()
+
+
+def test_segfault_on_realloc(tmpdir):
+    """ assert that we are able to reallocate var from un un-close creator #31 """
+    creator = Creator(str(tmpdir / "1.zim"), "welcome", "eng", 2048)
+    # following used to segfault
+    creator = Creator(str(tmpdir / "2.zim"), "welcome", "eng", 2048)
```
mgautierfr, line 149: It worth to mention that we don't really "reuse" a variable here.
The second `creator = Creator(...)` create a second `Creator` and associate the name "creator" to it. So the first `Creator` is associated with no name and so the gc may delete it (without calling `close`).

This issue should be throwable with :
```
creator = Creator(...)
del creator
```
(We can even force the delete of the unassociated object with `gc.collect()`)

However, this is just for information, it is not a blocker for this PR at all.

libzim/writer.py
rgaudin: You're right. My idea was to prevent users from getting a `RuntimeError` when calling `.close()` twice because the user has no way to know whether the c_creator has been finalized or not which is somewhat problematic.

But this is a different topic so I'm removing that from here and will open a separate request to get that information.

tests/test_libzim.py
rgaudin: Agrees, following the issue's example but it's not required. Fixed it

tests/test_libzim.py
```diff
@@ -159,3 +159,28 @@ def test_creator_params(tmpdir):
     assert zim.filename == path
     assert zim.main_page_url == main_page_url
     assert bytes(zim.get_article("/M/Language").content).decode("UTF-8") == index_language
+
+
+def test_segfault_on_realloc(tmpdir):
+    """ assert that we are able to reallocate var from un un-close creator #31 """
```
mgautierfr, line 165: This docstring is not totally exact now.

rgaudin:
> The final code is good.
> But 7 commits for this is a bit too much :)
> And please, rebase on `master` instead of merging `master` into the branch.

Yes, the *Resolve Conflict* button on github is a curse :) Kept the tests in a separate commit though.

mgautierfr:
We are good.

PR #47, opened 2020-06-16
Various improvements

### LINKED ISSUE #29: don't require min_chunk_size on Creator
Currently, `libzim.writer.Creator` requires all its arguments to be passed:

```py
def __init__(self, filename, main_page, index_language, min_chunk_size):
```

While it's understandable for `filename` and `main_page`, and debatable for `index_language`, `min_chunk_size` is definitely an advanced feature and should receive a default value.

### LINKED ISSUE #30: get_metadata() doesn't return a dict
`get_metadata()`'s docstring announces a returned `dict` with the file's metadata.

Instead, it expects a parameter, the metadata `Key` and returns a memory-view of the value for this key.

* we probably want `str` values (about all use cases) here as the raw content would still be available at `r.get_article("M/Tags").content`
* shall we have that `dict` of all defined meta somewhere? It would definitely be useful.
* docstring should be adjusted to the chosen behavior

### LINKED ISSUE #37: ReadArticle __repr__ is incorrect
```py
    def __repr__(self):
        return f"{self.__class__.__name__}(url={self.longurl}, title=)"
```

### LINKED ISSUE #38: get_article shouldn't raise RuntimeError
Currently, `ReadArticle.get_article()` raises a `RuntimeError` should the URL be incorrect. I believe we should either:

* Add an `.has_article(url)` methods that returns a boolean informing about the existence of the article
* Raise a custom, usable Exception like `ArticleNotFound` or something.

### LINKED ISSUE #45: should we accept pathlib.Path?
With the general move in the stdlib to accept `pathlib.Path` for all path-related operations, it seems a little odd and dated to *have to* to send `str` to the libzim.

I'd like to harmonize with this move so that the `Creator` and `File` can take both `str` and `Path` and we'd check the type in the python wrapper and converts it to `str` for the binding.

I shall mention that functions usually returns same type as provided. We only output path for `File.filename` I think. We would acknowledge that we'd still return `str` here (otherwise we'd have to store the type info in the wrapper)

rgaudin:
Here's a collection of individual small fixes for identified issues:

*  Fixed #29: using default values on Creator init 
* Fixed #37: fixed `__repr__` for ReadArticle and File, added on Article
* Fixed #38: raising NotFound on unavailable article.
this might not be wanted as per discussion on #38 and matthieu's suggestion to leave that to new libzim's. I believe this might be forward compatible but we can strip this out if we think it won't.
* Issue #38: added has_article(url) to test an URL.
Currently uses the `NotFound` exception but could be changed to use `RuntimeError` should we opt NotFound out.
* Fixed #30: get_metadata() now returns a dict of metadata
Didn't receive comment on that issue so I impl. what made sense to me.
* Fixed #45: Creator now accepts pathlib for filename
* 1e65ca5 simplifies the public Creator API by refactoring what's used only once and by using better docstrings and annotations. **warning**: added a `._closed` prop to avoid the double-close issue we'd get on exiting the contextmanager and (auto) deletion.

Please look at them individually and let me know which would make sense to push as separate PRs.

### DIFF
--- a/libzim/wrapper.pyx
+++ b/libzim/wrapper.pyx
@@ -31,6 +31,7 @@ from libcpp.memory cimport shared_ptr, make_shared, unique_ptr
 
 import datetime
 
+
 class NotFound(RuntimeError):
     pass
 
@@ -407,14 +408,13 @@ cdef class FilePy:
         return article
 
     def get_metadata(self, name):
-        """Get the file metadata.
+        """Get a metadata
         Returns
         -------
-        dict
-            A dictionary with the file metadata
+        bytes
+            Metadata article's content. Can be of any type.
         """
-        article = self.get_article(f"M/{name}")
-        return article.content
+        return bytes(self.get_article(f"M/{name}").content)
 
     def get_article_by_id(self, article_id):
         """Get a ZimFileArticle with a copy of the file article by article id.
--- a/libzim/wrapper.pyx
+++ b/libzim/wrapper.pyx
@@ -29,6 +29,7 @@ from libcpp.string cimport string
 from libcpp cimport bool
 from libcpp.memory cimport shared_ptr, make_shared, unique_ptr
 
+import pathlib
 import datetime
 
 
@@ -356,14 +357,14 @@ cdef class FilePy:
     ----------
     *c_file : File
         a pointer to a C++ File object
-    _filename : str
+    _filename : pathlib.Path
         the file name of the File Reader object
     """
 
     cdef wrapper.File *c_file
     cdef object _filename
 
-    def __cinit__(self, str filename):
+    def __cinit__(self, object filename):
         """Constructs a File from full zim file path.
         Parameters
         ----------
@@ -371,8 +372,8 @@ cdef class FilePy:
             Full path to a zim file
         """
 
-        self.c_file = new wrapper.File(filename.encode('UTF-8'))
-        self._filename = self.c_file.getFilename().decode("UTF-8", "strict")
+        self.c_file = new wrapper.File(str(filename).encode('UTF-8'))
+        self._filename = pathlib.Path(self.c_file.getFilename().decode("UTF-8", "strict"))
 
     def __dealloc__(self):
         if self.c_file != NULL:
--- a/libzim/wrapper.pyx
+++ b/libzim/wrapper.pyx
@@ -31,6 +31,9 @@ from libcpp.memory cimport shared_ptr, make_shared, unique_ptr
 
 import datetime
 
+class NotFound(RuntimeError):
+    pass
+
 #########################
 #         Blob          #
 #########################
@@ -392,13 +395,13 @@ cdef class FilePy:
             The Article object
         Raises
         ------
-            RuntimeError
+            NotFound
                 If an article with the provided long url is not found in the file
         """
         # Read to a zim::Article
         cdef wrapper.Article art = self.c_file.getArticleByUrl(url.encode('UTF-8'))
         if not art.good():
-            raise RuntimeError("Article not found for url")
+            raise NotFound("Article not found for url")
 
         article = ReadArticle.from_read_article(art)
         return article
@@ -413,12 +416,12 @@ cdef class FilePy:
         article = self.get_article(f"M/{name}")
         return article.content
 
-    def get_article_by_id(self, id):
+    def get_article_by_id(self, article_id):
         """Get a ZimFileArticle with a copy of the file article by article id.
 
         Parameters
         ----------
-        id : int
+        article_id : int
             The id of the article
         Returns
         -------
@@ -427,13 +430,15 @@ cdef class FilePy:
         Raises
         ------
             RuntimeError
+                If there is a problem in retrieving article (ex: id is out of bound)
+            NotFound
                 If an article with the provided id is not found in the file
         """
 
         # Read to a zim::Article
-        cdef wrapper.Article art = self.c_file.getArticle(<int> id)
+        cdef wrapper.Article art = self.c_file.getArticle(<int> article_id)
         if not art.good():
-            raise RuntimeError("Article not found for id")
+            raise NotFound("Article not found for id")
 
         article = ReadArticle.from_read_article(art)
         return article
--- a/libzim/wrapper.pyx
+++ b/libzim/wrapper.pyx
@@ -20,6 +20,7 @@
 
 cimport libzim.wrapper as wrapper
 
+from typing import Generator
 from cython.operator import dereference, preincrement
 from cpython.ref cimport PyObject
 from cpython.buffer cimport PyBUF_WRITABLE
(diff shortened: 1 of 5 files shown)


libzim/writer.py
```diff
@@ -135,7 +135,7 @@ class Creator:
         Zim file metadata
     """
 
-    def __init__(self, filename, main_page, index_language, min_chunk_size):
+    def __init__(self, filename, main_page="", index_language="eng", min_chunk_size=2048):
```
mgautierfr: I would not provide a default value for the main_page.
We need a value and there is no other way to set it than at `__init__` step.

tests/test_libzim.py
```diff
@@ -184,3 +184,11 @@ def test_double_close(tmpdir):
     creator.close()
     with pytest.raises(RuntimeError):
         creator.close()
+
+
+def test_default_creator_params(tmpdir):
+    """ ensure we can init a Creator without specifying all params """
+    creator = Creator(str(tmpdir / "test.zim"))
+    assert True  # we could init the Creator without specifying other params
```
mgautierfr, line 192: I think in is unnecessary. If you can't init the Creator, it will fail before.

libzim/wrapper.pyx
```diff
@@ -414,14 +417,19 @@ cdef class FilePy:
...
+                Or if the provided id is out of bound
         """
 
         # Read to a zim::Article
-        cdef wrapper.Article art = self.c_file.getArticle(<int> id)
+        cdef wrapper.Article art
+        try:
+            art = self.c_file.getArticle(<int> id)
+        except RuntimeError as exc:
+            raise NotFound(str(exc))
```
mgautierfr, line 454: If you have an exception here, it is about something else than a non existing article.
I would raise a `RuntimeError`.

libzim/wrapper.pyx
```diff
@@ -370,6 +370,18 @@ cdef class FilePy:
         """Get the filename of the File object"""
         return self._filename
 
+    def has_article(self, url):
```
mgautierfr, line 385: I'm not sure we need this method.

It may let think the check is less expensive than getting directly the article, but in fact this is the same.

As the user code will probably be :
```python
if zimfile.has_article("foo"):
    article = zimfile.get_article("foo")
else:
    print("Not found")
```

We will end to call twice `get_article`.

If there is no method :
```python
try:
    article = zimfile.get_article("foo")
except NotFound:
    print("Not found")
```

It will call the `get_article` only once and it follow the [Easier to ask for forgiveness than permission](https://docs.python.org/3/glossary.html#term-eafp ) principle.

(The EAFP principle doesn't forbid a `has_article` method, but as the implementation of `has_article` is a call to `get_article` I think we should not have it)

libzim/wrapper.pyx
```diff
@@ -406,15 +417,15 @@ cdef class FilePy:
...
         Returns
         -------
         dict
-            A dictionary with the file metadata
+            A dictionary with the standard metadata present on file
         """
-        article = self.get_article(f"M/{name}")
-        return article.content
+        return {name: bytes(self.get_article(f"M/{name}").content).decode("UTF-8")
+                for name in MANDATORY_METADATA_KEYS if self.has_article(f"M/{name}")}
```
mgautierfr, line 429: The new API will have a method called `get_metadata` that will return a specific metadata (given a name). It would be better to do the same here to avoid a new API change on python side.

libzim/writer.py
```diff
@@ -143,43 +184,68 @@ def __exit__(self, *args):
...
+        def write_metadata():
+            """Store _medtadata dict as MetadataArticle"""
+            for key, value in self._metadata.items():
+                if key == "Date" and isinstance(value, (datetime.date, datetime.datetime)):
+                    value = value.strftime("%Y-%m-%d")
+                article = MetadataArticle(key, value)
+                self._creatorWrapper.add_article(article)
+
+            article = MetadataArticle("Counter", get_counter_string())
+            self._creatorWrapper.add_article(article)
```
mgautierfr, line 242: I would not create a function just to call it after.
Simply write the code in the `close` body.

libzim/writer.py
rgaudin: OK, so the default in the wrapper is just for technical reason then. I get it.

tests/test_libzim.py
rgaudin: Yes, I just though it looked unfinished without it :) docstring makes it clear anyway.

libzim/wrapper.pyx
rgaudin: yeah you mentioned that would be revamped so was hoping we'd get something similar then. Will go away for now.

libzim/wrapper.pyx
rgaudin: It's usually because of an out-of-bound id. We can raise `RuntimeError` for that.

libzim/wrapper.pyx
rgaudin: Ah, good to know! Should we return the decoded string directly here? Otherwise it's quite useless compared to `get_article()`…
Also, should we provide this dict under another name?

libzim/writer.py
rgaudin: works for me

libzim/wrapper.pyx
mgautierfr: It will be not possible to get metadata with `get_article` (or equivalent in new API).
I'm not sure that we need to decode the string all the time. From https://www.openzim.org/wiki/Metadata, all metadata in `M` namespace are strings so we could decode it. But the favicon, even if it is in `-` namespace (and will probably move in `M`), is somehow a metadata. A call to `get_metadata('favicon')` should work and return a bytes.

> Also, should we provide this dict under another name?

I'm a bit disturbed with this dictionary. There are some mandatory keys in the specs but the implementation never enforce this. If we go this way, it should be in libzim itself, not in the wrapper. And if we don't, I'm not sure the wrapper should enforce this.

I don't know what is the best here (about encoding and dict). But we should think about this :)

rgaudin: OK, I am also quite uncomfortable with the `MANDATORY_METADATA_KEYS` and `Creator.mandatory_metadata_ok`. Can we agree this is a wrong interpretation of the spec (nothing's actually mandatory) and remove both?

Thanks for the additional clarification on future lib's `get_metadata()`.
I think it's a bit cumbersome to have to convert to bytes then decode manually to get a key/value string but I don't like magic decoding so I think we should just mirror get_article for now (and thus revert and fix the docstring as impl. was ok).

But this highlights the need to query existing metadata and there is no way in the current implementation to get the list of M/ articles. I can get the number of articles but that's it. I can not discover metadata and that's problematic.
Is there a way in the libzim to do it that's not exposed in this impl? The only way I see for now is to decrement-loop on article count (because M/ are stored before X/) and test urls.

mgautierfr: > OK, I am also quite uncomfortable with the MANDATORY_METADATA_KEYS and Creator.mandatory_metadata_ok. Can we agree this is a wrong interpretation of the spec (nothing's actually mandatory) and remove both?

There are some mandatory metadata in (https://www.openzim.org/wiki/Metadata). It was somehow a grey area between openzim and kiwix specification. But whatever how we interpret it, it must be consistent between c++ and python. I think we can remove it from python wrapper. If we need it, we must add it on c++ side.

> I think it's a bit cumbersome to have to convert to bytes then decode manually to get a key/value string but I don't like magic decoding so I think we should just mirror get_article for now (and thus revert and fix the docstring as impl. was ok).

We could at least convert it to `bytes` (instead of `memoryview`) as we know that the size is reasonable.

> But this highlights the need to query existing metadata and there is no way in the current implementation to get the list of M/ articles. I can get the number of articles but that's it. I can not discover metadata and that's problematic.
Is there a way in the libzim to do it that's not exposed in this impl? The only way I see for now is to decrement-loop on article count (because M/ are stored before X/) and test urls.

You are right. I haven't think about that but we definitively need to make the metadata discoverable. I will take this into account in the new API.

rgaudin:
@mgautierfr , all your comments have been addressed. There isn't much actual code
change beside moving blocks around in the whole PR so I kept it all here.

You can look at individual commits for a quick lookup.

What's changed since last review:
- `main_page` is now mandatory on Creator
- test on default params for Creator don't inclue dummy `assert True`
- `get_article_by_id()` now raises `RuntimeError` on out-of-bound – actually anything but `not art.is_good()`, the try block have been removed.
- `get_article_by_id()`'s `id` param was renamed `article_id` to not use reserved keyword
- `has_article()` method have been removed
- `get_metadata()` returns bytes of content of a single metadata article and not the discussed dict.
- all references to mandatory metadata are gone
- Creator former methods which had been moved inside `close()` (as used there and once) have now been put inline, still in that close() method.
- All docstrings were fixed for the reader API
- added `__enter__` and `__exit__` (does nothing) on reader so it can be used as a context manager (syntax sugar only)

rgaudin:
OK thanks ; will do soon

rgaudin:
Sorry about the delay, was a bit of a pain to reconcile but it's now clean and ready.

mgautierfr:
I'm sorry but it is not totally clean yet :)
Commit [`ed2267b` (#47)](https://github.com/openzim/python-libzim/pull/47/commits/ed2267bf7fac8bb041b280901afe66a6fe9164ff) remove the `str` in the test, but the acceptance of pathlib as argument come in the commit just after (https://github.com/openzim/python-libzim/pull/47/commits/5ecfab69ffe0586899cbd4ec8a3266c40ea88014)

About the execption, maybe we could return `KeyError` and `IndexError` instead of `NotFound` or `RuntimeError`. But we could change this in another PR, this one is long enough.

rgaudin:
That should be good now. Rebased on master. Tests pass on every single commit.

mgautierfr:
We are good !

PR #49, opened 2020-06-19
rename get_namespaces_count with get_namespace_count

rgaudin:
As this method returns the count for a single namespace, its name should be singular


rgaudin:
rebased

PR #50, opened 2020-06-23
Upgrade cython to latest stable

rgaudin:
Cython is a fast evolving project with frequent fixes and improvements.
Keeping up-to-date on stable version makes sense as a failure to build
with a newer Cython would indicate a problem on our codebase.

@afreydev, what was the reason for using such an old version? `0.29.6` is dated Feb 27, 2019.


afreydev:
At that moment, there were a lot of problems related to segmentation faults. This version doesn't cause those problems. That's all. You can try upgrade it.

rgaudin:
rebased

PR #55, opened 2020-06-26
Forward exceptions from Article's virtual methods

rgaudin:
This addresses (badly I presume) the problem that users need to stop on user-code error inside `Article`'s virtual method.
It's essential otherwise they'll endup with garbaged ZIM files.
This first commit (8ed926d) achieves that but there is probably a be a better way to do it.

--

Another related commit regards the *next step* in this process, being able to stop the creation process when an error occurs.
The only workaround I found was to prevent calling `finalize()` (node-libzim and zimwritefs exits before calling it for instance).
As we call `close()` and thus `finalize()` on `__exit__` and `__del__`, for convenience, we have to bypass this.

It works fine in scrapers but not on the tests as the reader's tests trigger a manual garbage collection which the libzim is not happy with, as `finalize()` was never called…

To overcome that yet still test this important *capability*, I'm instanciating a new VM in the test.

@mgautierfr, please consider it a POC to start the discussion on this. We can surely opt this test out for now . Exception stuff is mandatory though.

### DIFF
--- a/libzim/lib.cxx
+++ b/libzim/lib.cxx
@@ -65,8 +65,10 @@ std::string ZimArticleWrapper::callCythonReturnString(std::string methodName) co
     int error;
 
     std::string ret_val = string_cy_call_fct(this->m_obj, methodName, &error);
-    if (error)
+    if (error && error == 1)
         throw std::runtime_error("The pure virtual function " + methodName + " must be override");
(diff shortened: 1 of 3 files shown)


rgaudin:
Superseeded by #56.

PR #60, opened 2020-07-01
Use SPDX license identifier in setup.py

legoktm:
See <https://spdx.org/licenses/>.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
     },
     author="Monadical Inc.",
     author_email="jdc@monadical.com",
-    license="GPLv3+",
+    license="GPL-3.0-or-later",
     description="A python-facing API for creating and interacting with ZIM files",
     long_description=get_long_description(),
     long_description_content_type="text/markdown",
@@ -132,7 +132,7 @@ setup(
         "Programming Language :: Python :: 3.8",
         # "Typing :: Typed",
 
-        "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
+        "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
         "Natural Language :: English",
         "Operating System :: OS Independent",
     ],


rgaudin:
I think it would have been wise to share some information on this. I am not familiar with SPDX and had to dig that.

What's the rationale for the change? I see that there is [no PEP yet](https://github.com/pombredanne/spdx-pypi-pep/pull/2/files) and that details [are still being discussed now](https://discuss.python.org/t/improving-license-clarity-with-better-package-metadata/2154/75) regarding how to actually express this.

So my questions are:
* Is PyPi recognizing it (it's free text but we're using a string that's recognized) now?
* What's the benefit of doing that?

legoktm:
> I think it would have been wise to share some information on this. I am not familiar with SPDX and had to dig that.
> ...
> So my questions are:
> 
>  * Is PyPi recognizing it (it's free text but we're using a string that's recognized) now?
> 
>  * What's the benefit of doing that?

Sorry about that. In short, SPDX is a standardized way of referring to free software licenses using short IDs to make it easier for automated tools (and honestly humans too) to determine what the intended license is.

Also, while it isn't a problem in this case, I've run into enough cases where people forgot the `+` to indicate "or later" in the license. The FSF [also prefers using `-or-later`](https://www.gnu.org/licenses/identify-licenses-clearly.html).

For me, I use various automated tools (e.g. [licensecheck](https://manpages.debian.org/buster/licensecheck/licensecheck.1p.en.html)) when reviewing packages for license compatibility while packaging them in Debian. Using SPDX identifiers everywhere just makes it more likely that those tools get it right.

Even if PyPI hasn't adopted/recommended SPDX identifiers yet, I think it's still a good idea, and likely to happen eventually anyways, given that everyone else (composer, npm, cargo, ...) has adopted them.

rgaudin:
OK, then maybe fix the license classifier in same commit 

```
License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
```

legoktm:
Done

PR #67, opened 2020-07-07
Added tests to ensure Article's methods all raises if not overriden

rgaudin:


mgautierfr:
This seems a bit too complex. Especially, you are testing that not implementing a method correctly raise an exception by implementing the method and calling the "super" implementation.

Pytest has a fixture to monkeypath a object (and remove a method/attribute) https://docs.pytest.org/en/stable/monkeypatch.html
It is probably better to use it to remove method from a correctly implemented article.

And is is better to move the method to remove to a parametrized fixture (https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures) instead of having a loop in the test.

rgaudin:
You're right it's clearer. @mgautierfr can you take another look?

tests/test_libzim.py
```diff
@@ -274,3 +275,58 @@ def get_redirect_url(self):
...
         assert reader.get_article(redirect_url).get_redirect_article().longurl == url
+
+
+@pytest.mark.parametrize(
+    "no_method", [m for m in dir(Article) if m.split("_", 1)[0] in ("get", "is", "should")],
+)
+def test_article_overriding_required(tmpdir, monkeypatch, no_method):
+    """ ensure we raise properly on not-implemented methods of Article """
+
+    class OverridenArticle(Article):
```
mgautierfr, line 286: We can move the `OverridenArticle` outside of the test and have the `no_method` fixture loop over the method of `OverridenArticle` instead of `Article`.

tests/test_libzim.py
```diff
@@ -274,3 +275,58 @@ def get_redirect_url(self):
     with File(path) as reader:
         assert reader.get_article(redirect_url).is_redirect
         assert reader.get_article(redirect_url).get_redirect_article().longurl == url
+
+
+@pytest.mark.parametrize(
+    "no_method", [m for m in dir(Article) if m.split("_", 1)[0] in ("get", "is", "should")],
+)
```
mgautierfr, line 282: I didn't know that we could directly parametrize a test. I was using to parametrize a fixture and use the fixture in the test. Good to know :)

tests/test_libzim.py
rgaudin: Done.

PR #69, opened 2020-07-08
Include tests/*.zim files in the release tarball

### LINKED ISSUE #68: tests/*.zim files not included in release tarball
If you download https://files.pythonhosted.org/packages/26/8e/201f1ed560f83f5fd4a87fe3ec52960d078f1a31c8f1670ac16cb92fe3a3/libzim-0.0.3.post0.tar.gz it's missing the two zim files in tests/ so the tests fail. I believe we need a MANIFEST.in file for this.

Also, from a licensing standpoint, it would be nice if we could use zim files with fully public domain content so then I don't have to keep track of the specific copyright status for Debian.

legoktm:
This will allow tests to pass from people/distributions (e.g. Debian) using
the release tarball.

Fixes #68.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 include LICENSE
 include README.md
 include tests/*.py
+include tests/*.zim
 include pyproject.toml
 
 recursive-include libzim *


legoktm:
Superseded by #76.

PR #75, opened 2020-07-17
Ability to choose compression algoritm

rgaudin:
libzim 6.1.8 added the ability to choose the compression type on creation.
This adds support for it in the Creator and other required parts.

On user API, the main change is a `compression` argument to the Creator.
This can receive either a string (`default`, `none`, `zip`, `lzma`, `zstd`) or a member
of the new `Compression` enum (members named after strings above)

This enum converts to `zim::CompressionType` and that is forwarded down to the libzim.

## Breaking change

I chose to add the new `compression` argument to the Creator before the `min_chunk_size` as it seems there might better chances people would customize it than the chunks size.
I might be wrong or it might just not be wanted. Let me know and I'd change that. Because of this change, two tests using positional arguments were changed.

--

**Note**: this assumes https://github.com/openzim/libzim/issues/371 is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.0.4
 
+* added compression argument to Creator to set compression algorithm (libzim.writer.Compression)
+* Creator positional arguments order changed: min_chunk_size moved after compression
 * using libzim 6.1.8
 
 ## 0.0.3.post0
--- a/libzim/lib.cxx
+++ b/libzim/lib.cxx
@@ -204,8 +204,8 @@ std::string ZimArticleWrapper::getNextCategory()
 class OverriddenZimCreator : public zim::writer::Creator
 {
 public:
-    OverriddenZimCreator(std::string mainPage)
-        : zim::writer::Creator(true),
+    OverriddenZimCreator(std::string mainPage, zim::CompressionType compression)
+        : zim::writer::Creator(true, compression),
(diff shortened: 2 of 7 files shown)


mgautierfr:
We already discuss a bit in https://github.com/openzim/libzim/issues/371 but I think that the wrapper should simply don't provide the "zimCompDefault" value. (The next libzim api will probably change in the same time than libzim2).

On the `min_chunk_size`, very few (if not none) user will set this. We could move in a "configuration" method on the creator instead of having it in the constructor (as it is made on libzim side).

rgaudin:
@mgautierfr, I've removed `zimcompDefault` and now defaults to lzma.

As per `min_chunk_size`, the Creator now has a `configuration` property returning a dict with only `min_chunk_size` at the moment. Let me know if that's how you want it.

mgautierfr:
I'm not sure a user can change the `min_chunk_size` the way you've done it.

I was thinking about a simple `set_min_chunk_size` method (or a propery) on the Creator, calling the cpp `setMinChunkSize`. But if it is called after `startZimCreation` it is useless.

It appears that the simpler is to keep the `min_chunk_size` argument in the `__init__` after all.

rgaudin:
Ah ok, didn't realize you meant post-init. OK, dropping that commit then.

rgaudin:
ready

PR #78, opened 2020-08-12
Require get_size() implementation by user

rgaudin:
`getSize` being implemented in the wrapper, and its implementation relying on returning
the size of get_data(), it prevented the use of get_filename (allowed by libzim).

This now requires user to implement `get_size()`, not making any assumption about it.

- API break: `get_size()` now mandatory on Article
- WritingBlob now exposes a `size()` that returns the C Blob's size so we can still fetch
the size from `get_data()`
- tests checking for implementation of mandatory methods adapted so get_size returns a positive
size when testing get_data (otherwise it's skipped)
- Added a test for `get_filename()` usage


rgaudin:
rebased

Issue #42, opened 2020-06-10
Exception inside contextmanager should cancel the zim creation


rgaudin:
When using the context-manager, should a (non libzim) error occur, the exception is raised but the finalization is done on the Creator as if everything went well.

```py
with libzim.writer.Creator(
    "test_x07.zim", main_page="A/index.html", index_language="eng", min_chunk_size=2048,
) as zfile:
    zfile.add_article(DumbArticle("index.html", "hello", ARTICLE_MIME, "bonjour"))
    raise Exception("outch")
    zfile.add_article(DumbArticle("page2.html", "hello2", ARTICLE_MIME, "bonjour2"))

T:0; A:4; RA:0; CA:4; UA:0; FA:0; IA:1; C:0; CC:0; UC:0; WC:1
T:0; Waiting for workers
T:0; ResolveRedirectIndexes
Resolve redirect
T:0; Set article indexes
set index
T:0; Resolve mimetype
T:0; create title index
T:0; 6 title index created
T:0; 2 clusters created
T:0; write zimfile :
T:0;  write mimetype list
T:0;  write directory entries
T:0;  write url prt list
T:0;  write title index
T:0;  write cluster offset list
T:0;  write header
T:0;  write checksum
T:0; rename tmpfile to final one.
T:0; finish
Traceback (most recent call last):
  File "./demo.py", line 55, in <module>
    raise Exception("outch")
Exception: outch
```

This results in a valid ZIM file on the filesystem but lacking the second article of course.

I think the expected behavior would be to *cancel* the ZIM creation and remove temporary files.

@mgautierfr @kelson42 ?

rgaudin:
@mgautierfr, I tested the `thread_exception` branch on libzim but I see no difference in how an exception in a ContentProvider is handled"

```py
def test_creator_exception_item(fpath):
    class AContentProvider:
        def get_size(self):
            return 1

        def feed(self):
            raise FileNotFoundError("missing file")

    class AnItemWithCP(StaticItem):
        def get_contentprovider(self):
            return AContentProvider()

    with Creator(fpath) as c:
        with pytest.raises(RuntimeError, match="FileNotFoundError"):
            c.add_item(AnItemWithCP())
````

It crashes with an output similar to:

```
Resolve redirect
set index
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Traceback (most recent call last):
  File "libzim/wrapper.pyx", line 127, in libzim.wrapper.blob_cy_call_fct
[...]

PR #82, opened 2020-09-21
Libzim next

mgautierfr:
This is an adaptation of python-libzim to the new libzim api.

The new api concentrate on reading and writing part (not search).
- It add a intermediate step to access the content (Archive->Entry->Item).
- On writing part, it is no more necessary to inherit the creator. But it is still the case for `Item`
(and `ContentProvider`).
Metadata and redirections are adding directly using `Creator` specific method instead of inherit a `Item`.

C++ api provides some small default implementation for item and contentprovider.
It is not wrapped in python api. This is open to discussion but I think we should not wrap them but reimplement them in python (still being part of the api).
Wrapping the default implementation may be difficult (especially the fact that a `Item` may return a python or cpp implemented `ContentProvider`). The implementation is pretty simple, we can directly implement them in python without having to take care of that.


rgaudin:
**ISSUE**: unexpected suggest behavior

Using libzim_next (nightly version from 2021-01-18)

Using [isago ZIM](https://mirror.download.kiwix.org/zim/.hidden/dev/isago.ml_fr_all.zim), which is composed of a few articles with all the same title (`I SAGO – tarifs de l'éléctricité au Mali`), I get a correct behavior when **searching**:

- searching for `recharge` gives me `A/isago.ml/faq` and `A/isago.ml/conseils`
- searcing for `montant` gives me `A/isago.ml/` and `A/isago.ml/conseils`

All of this is OK.

When getting suggestion, given I have 4 articles with same title I should either get no result or 4 results.

- suggest for `recharge` -> no result. That's OK
- suggest for `sago` gives me `A/isago.ml/`. That's where I would expect to receive the four URLs.

Might be a libzim error or just a misinterpretation from my side.

rgaudin:
**ISSUE**: Segfaults when setting config

```py
import pathlib
from libzim.writer import Creator

with Creator(pathlib.Path("test.zim")) as c:
   c.configIndexing(True, "eng")
```

This segfaults while the same code without `configIndexing` created a proper (yet empty) ZIM.

rgaudin:
**QUESTION**: Should we provide access to entry ID?

Entry's ID is not exposed in the python API and I don't think we need it but the fact that `entry.get_redirect_entry()` would raise `RuntimeError: Entry 2 is not a redirect entry.` makes me think this is not coherent.
Either we should provide this ID or be path-first and print that libzim error message with the path instead of the ID.

rgaudin:
**QUESTION**:  can you clarify the difference between the Entry's path and the Item's path?

Are those those separate values in the ZIM or is the Entry's path just a shortcut to the item's one?
It would seem so from a writer's perspective as we only create *Item* and thus provide this data only once
but we can have something like

```py
zim.main_entry.path
> 'mainPage'

zim.main_entry.get_item().path
> 'home'
```

Is this just because `get_item()` resolves the redirect?

rgaudin:
**QUESTION**: Main Entry's behavior between new and old ZIMS

With a ZIM created using old libzim, it gets the actual home article on `zim.get_main_entry`.
Using this new creator, I can only get a stub redirect

``` py
# using older zim
zim.main_entry
> Entry(url=A/index.html, title=index.html)


# using new creator
zim.main_entry
> Entry(url=mainPage, title=mainPage)
```

Is this the wanted behavior? I think we'll have to adapt scraper/tools' tests to this.

rgaudin:
**QUESTION**: Why are we returning `self` on `configXXX()` calls?

rgaudin:
**ISSUE**: Interpreter crash on missing `gen_blob()`

If the subclassed ContentProvider doesn't implement `gen_blob`, the interpreter crashes with `Abort trap 6`. See `test_notimplementing_contentprovider_gen_blob` in `test_libzim.py`

rgaudin:
**QUESTION**: Should we provide start/end methods?

We've discussed that in the previous impl. I believe. I don't have a clear use case for this but It would be close to nothing to maintain and would add some flexibility to users.

rgaudin:
**QUESTION**: Am I right to assume that kiwix-serve will not be serving metadata on `M/` prefix with that new libzim (unless it implements a new, special behavior for it) ?

Related: with this new impl, we don't have access to other namespaces content (except metadata through those 2 methods), right ?

rgaudin:
> Related: with this new impl, we don't have access to other namespaces content (except metadata through those 2 methods), right ?


Please disregard this comment, I understand we can access some namespaces by prefixing the path.

> There is no notion of namespace in the API. If you want a metadata, use Archive::getMetadata(). If you want an entry, use Archive::getEntryByPath(). Either by using a "full" path (with the namespace) or a "simple" path (for compatibility libzim search in namespace 'A', 'I', 'J' and '-').

mgautierfr:
Here few answers.
Please open issues for the bugs.

---

> ISSUE: unexpected suggest behavior

This is a real issue with the search but probably not related to libzim next.
The suggestion is base on a xapian database feeded only with title. The results come from xapian. We probably miss creating the database or something else.

> ISSUE: Segfaults when setting config

This is a issue. Does it crash directly when you call the configIndex method or later ?
(With the code you've provided, "later" is just after as we are closing the contextManager)

> QUESTION: Should we provide access to entry ID?

You are right. We should only use path.

> QUESTION: can you clarify the difference between the Entry's path and the Item's path?

On a zim structure point-of-vue, there is no Entry or Item.
There are dirents, and dirents may be a redirect one or a "classic" one (with content associated).

The new API hides this and create two new entities that are purely conceptual.
The entry is the entry point to something. It could be a redirect or a classical dirent.
The item is the "content". It units the classical dirent and the associated content.

The item is not a different thing than a entry. It is a "upgraded" entry.
The path/title of a entry is the same than the associated item.

`get_item()` resolve the redirect for you, so potentially the given item is not the "same" than the entry.
Cpp `getItem()` has a boolean parameter telling if we should resolve the redirect and give the resolved item or not (and raise a exception if the entry is a redirect)
I don't remember why I've not exposed this parameter in the python version.

On the writer side, you can indeed only add a direction (using `addRedirection`) or add an item (`addItem`). 

> QUESTION: Main Entry's behavior between new and old ZIMS

The "new" mainPage is a redirection to the entry configured with `setMainPath` (before it was directly the configured entry).
This is needed as on kiwix-serve we need to have a redirection to ensure internal relative path are correct.
We were handle this in kiwix-lib but it is better to have it directly in libzim.


`zim.get_entry_by_path(zim.main_entry.get_item().path)` should always work.

> QUESTION: Why are we returning self on configXXX() calls?

This is useful to chain calls :
```python
zim_creator = Creator(tmpdir / "test.zim")
zim_creator.configIndexing(True, "eng").configFoo().configBar()
```

One important point slightly explained in cpp (https://github.com/openzim/libzim/blob/master/include/zim/writer/creator.h#L35-L46) and not explain in python is that the
`configXXX` must be called **before** the start of the creation process (where `setXXX` can be call before or after the start)
It may explain the segfaults with `configIndex`.
(We should document that, and properly handle the error if user call `configXXX` after the creation)

> ISSUE: Interpreter crash on missing gen_blob()

I will investigate.

> QUESTION: Should we provide start/end methods?

Not sure of that. It is easier to add them if needed than remove them. Let's move without them for now.

> QUESTION: Am I right to assume that kiwix-serve will not be serving metadata on M/ prefix with that new libzim (unless it implements a new, special behavior for it) ?

Yes, `get_entry_by_path` only allow to get content add in the zim using `addRedirection` or `addItem` other kind of data must use another (specific) methods.
kiwix-serve already has a way to get metadata for a zim file : `/meta?content=foo_zim_name&name=author`

> I understand we can access some namespaces by prefixing the path.

This works only for old zim file for compatibility.
On new zim file, the content is always search in the "main" namespace. You cannot access `M`/`W`/`X` namespace with `get_entry_by_path`

rgaudin:
Thanks for the detailed response, here are the remaining actions to take:

- [ ] Find appropriate ticket on libzim regarding suggestions and add this use case as comment @rgaudin 
- [ ] Properly raise on `configXXX()` being called after start @rgaudin 
- [ ] Open a ticket on libzim to fix error message (use path instead of ID) @rgaudin 
- [ ] Implement `FileProvider` and `StringProvider` in the wrapper (in python) @rgaudin 
- [ ] Document new API and specificity (README) @rgaudin 
- [ ] Open tickets here for implementing `getIndexData` and `getHints` @rgaudin 
- [ ] Add `follow=True` param to `get_item` @rgaudin (unless that proves problematic)
- [ ] Fix Interpreter crash on missing gen_blob() @mgautierfr

rgaudin:
@mgautierfr, I've added support for a few methods on the reader that I believe are useful.
I've added those on separate commits so we can exclude them if you think we don't want pylibzim users to access them.
I think all of those make sense except maybe for `Item.size` which is a libzim-way around `Item.content.nbytes`
I've voluntarily skipped `getEntryByTitle` and similar though.

* `Archive.favicon_entry`
* `Archive.uuid`
* `Archive.new_namespace_scheme`
* `Archive.has_fulltext_index` and `Archive.has_title_index`
* `Archive.filesize`
* `Archive.metadata_keys`
* `Entry.index` and `Item.index`
* `Item.size`

libzim/wrapper.pyx
```diff
@@ -390,68 +390,72 @@ cdef class Item:
...
     def __dealloc__(self):
         if self.c_archive != NULL:
             del self.c_archive
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        pass
+
```
mgautierfr: I'm not sure we want this.
The context manager implies a under some action are valid. When the context exit, you cannot do some stuff.

This is needed when creating a zim file (we have a context of creation, once it is ended, we cannot add more item).
This is useful with "plain" file (where we want to close the file at the end of the context)

But here, the context represent "nothing". You can open a archive, get a item, close the archive and still use the item after.
And we do nothing when we enter or close the context.

libzim/wrapper.pyx
```diff
@@ -197,29 +197,39 @@ cdef class Creator:
...
+            raise RuntimeError("ZimCreator started")
         self.c_creator.configNbWorkers(nbWorkers)
         return self
 
-    def setMainPath(self, str mainPath) -> Creator:
+    def setMainPath(self, str mainPath):
         self.c_creator.setMainPath(mainPath.encode('utf8'))
 
-    def setFaviconPath(self, str faviconPath) -> Creator:
+    def setFaviconPath(self, str faviconPath):
```
mgautierfr, line 233: It would be better to `return self` instead of removing the `-> Creator`.
But it is not really important.

libzim/writer.py
```diff
@@ -109,6 +109,7 @@ def configCompression(self, compression: Compression):
         if not isinstance(compression, Compression):
             compression = getattr(Compression, compression.lower())
         super().configCompression(compression)
+        return self
```
mgautierfr, line 146: `return super().configCompression(compression)`

libzim/writer.py
```diff
@@ -70,6 +71,39 @@ def gen_blob(self):
...
+
+    def get_size(self):
+        return self.size
+
+    def gen_blob(self):
+        bsize = 1048576  # 1MiB chunk
+        res = self.filehandler.read(bsize)
+        while res:
+            yield Blob(res)
+            res = self.filehandler.read(bsize)
```
mgautierfr, line 100: We can use a context manager here :
```
bsize = 1048576
with open(self.filepath, "rb") as f:
    res = f.read(bsize)
    while res:
        yield Blob(res)
        res = f.read(bsize) 
```
This way we don't have to store the filehandler (and close it in __del__).

tests/test_libzim.py
```diff
@@ -437,3 +443,155 @@ def get_contentProvider(self):
...
+
+        def get_mimetype(self):
+            return "application/octet-stream"
+
+        def get_contentProvider(self):
+            return self.cp
+
+    with Creator(zim_path) as zim_creator:
+        with pytest.raises(IOError):
+            zim_creator.add_item(FileItem(path, fpath))
```
mgautierfr, line 597: Well, if we move the gen_blob using a context manager, we will catch the error when libzim want to get the content.
By opening it the constructor, we catch the error when we create the `FileProvile`...

mgautierfr (commented):
> I think all of those make sense except maybe for Item.size which is a libzim-way around Item.content.nbytes

`Item.size` is good. It avoid getting the data to just have the nbytes (even if `content` is a memoryview)

> Entry.index and Item.index

I'm not sure we want this. There is no api in libzim taking a index. If I remember correctly I'v added the `getIndex` method only for unit test purpose.

> And please take a look at 3658ab0, there's probably a better way than using a temp string.

Without adding a api on c++ uuid to directly create a `std::string` I don't see a better way.

libzim/wrapper.pyx
rgaudin: Harmonized with the assumption that we'd return self on those that *have to* be called before starting/entering. setFaviconPath and setMainPath can be called after start so I harmonized this way.
Should we return self here, I suggest we do the same for `setMainPath`.

We should probably add some docstrings about pre-start usage requirement for config* as well.

tests/test_libzim.py
rgaudin: No, the error is raised by the getsize call in the constructor, regardless of where the file is opened.

libzim/wrapper.pyx
rgaudin: returned `self` on both

rgaudin (approved):

mgautierfr:
> > ISSUE: Interpreter crash on missing gen_blob()
> I will investigate.

Can you test with https://github.com/openzim/libzim/tree/thread_exception branch ?

rgaudin:
- renamed methods and props to lower_case
- rewrote all tests: cleaner and easier to maintain
- removed tests zims from repo: using libzim ones (dowloaded on test session start). will need to check with @legoktm if this is a problem for debian or not.
- changed nightly to `2021-01-27` so we have zstd by default (and can test it)
- some cleanup on writer:
  - setting `_started` in init
  - removed `bool_cy_call_fct` as it's not used anymore
  - added a `__init__()` on Creator so it can be subclassed
- made id/index in reader *private* (prefixed)
- exposing Entry and Item on reader otherwise not reachable (not useful for common use case though)
- eventually passed setup.py through black/flake and used black's default line-length as only a few unimportant bits were not compliant

Coverage is not at 100% but close:

- I'm not sure how to test those edge cases in `ReadingBlob`.
- I failed to get coverage to ack the `blob_cy_call_fct()` block although it does run it. if you have any clue…

I am ready to merge this as soon as @mgautierfr agrees

libzim/wrapper.pyx
```diff
@@ -164,348 +177,431 @@ cdef class Creator:
...
+
+    def __enter__(self):
+        cdef string _path = str(self._filename).encode('utf8')
+        with nogil:
+            self.c_creator.startZimCreation(_path)
+        self._started = True
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if True or exc_type is None:
```
rgaudin, line 293: https://github.com/openzim/python-libzim/issues/42

mgautierfr:
I agree to merge. There is probably more to do but let's do it in other PRs.

PR #103, opened 2021-03-16
Fixed suggestion count estimation

rgaudin:
Following discussion at openzim/libzim#521, it appears
pylibzim is returning an invalid estimated number of suggestions while libzim is correct
Adjusted the search object's range to match what is done in libzim.

Also updated the libzim version with test against to latest nightly as it includes related changes. Thus had to update the filesize of the blank zim file we create in tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 on: [push]
 
 env:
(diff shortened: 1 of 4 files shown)


rgaudin:
@mgautierfr can you take a look? tests are broken on master as we use an older nightly that's not available anymore (we only keep last 31)

PR #115, opened 2021-08-03
Add support for Hints in python-libzim

mgautierfr:
This PR is not fully tested, no check that created zim are correct.
The error handling is not perfect. No handling of :
- get_hints returning something else than a dict.
- get_hints returning a dict where the key is something else that a Hints.
- ....


kelson42:
@mgautierfr Does that fix https://github.com/openzim/libzim/issues/590 as well?

mgautierfr:
> @mgautierfr Does that fix openzim/libzim#590 as well?

Not entirely. It allow sotoki to be updated to pass the correct hints to libzim (which will fix the issue)

kelson42:
@mgautierfr hmmm... then it is not clear what is left to do to close the libzim bug.

rgaudin:
> There is no libzim bug. Only a wrong usage of it. we must update its user (python-libzim and sotoki) to correctly use it.
> 
> I've made half of the work here and @rgaudin seems to have made the other half as he has generated zim file with correct article listing ([kiwix/kiwix-js#708 (comment)](https://github.com/kiwix/kiwix-js/issues/708#issuecomment-892052042))
> 
> What we need to do is stabilize this API, be more resilient to errors and maybe add some tests (This PR already update tests to pass hints, but do not test more)

I'm on it

rgaudin (approved):
Fine with me. Used it with sotoki.

Added tests in 6528b53. @mgautierfr please check whether you are fine with this approach or not

libzim/wrapper.pyx
```diff
@@ -179,8 +179,8 @@ cdef public api:
         cdef map[HintKeys, uint64_t] ret;
         try:
             func = getattr(obj, method.decode('UTF-8'))
-            hintsDict = func()
-            return convertToCppHints(func())
+            hintsDict = {k: pybool(v) for k, v in func().items() if isinstance(k, Hint)}
```
mgautierfr, line 182: Hints values are unsigned integer. They can be something else than a bool.
Future hints may be some kind of group that libzim will use to regroup item in same cluster.

For now we are only use bool so we are ok, but we may have to change this.

tests/test_libzim_creator.py
```diff
@@ -555,6 +555,76 @@ def get_hints(self):
...
+            path="4", title="", targetPath="0", hints={Hint.COMPRESS: True}
+        )
+
+        # non-existent Hint
+        with pytest.raises(AttributeError, match="YOLO"):
+            c.add_item(StaticItem(path="0", title="", hints={Hint.YOLO: True}))
+
+        with pytest.raises(AttributeError, match="YOLO"):
+            c.add_redirection(
+                path="5", title="", target_path="0", hints={Hint.YOLO: True}
```
mgautierfr, line 624: Can you test with other keys ? (`5`, `"yolo"`, ...)

rgaudin: Yep. Adding that. should be filter. Should definitely add a `"FRONT_ARTICLE"` (`str`) key test as well

libzim/wrapper.pyx
mgautierfr: Ideally we should have cast specific to the key used. But it would always be bool for now.

rgaudin: OK, so should we keep it as it is for now and defer that change to whenever we need it ? ie. in a future libzim release with additional hints?

mgautierfr: yes

rgaudin: 👍

tests/test_libzim_creator.py
```diff
@@ -614,6 +614,10 @@ def test_hints_values(fpath):
         c.add_redirection(
             path="4", title="", targetPath="0", hints={Hint.COMPRESS: True}
         )
+        # filtered-out values
+        c.add_item(StaticItem(path="5", title="", hints={5: True}))
```
mgautierfr, line 618: Would be nice to also test with the int value corresponding to the real value of the hint key.
(Should be accept it or filter it ?)

mgautierfr (commented):
Few comments on the tests.
(I cannot approve or request changes on the PR as I'm the one who open it)

mgautierfr:
(The push-force only rewrite my commit message to remove the WIP)

tests/test_libzim_creator.py
```diff
@@ -555,6 +555,76 @@ def get_hints(self):
...
+def test_nondict_hints(fpath):
+    class AnItem:
+        def get_path(self):
+            return ""
+
+        def get_title(self):
+            return ""
+
+        def get_mimetype(self):
+            return ""
```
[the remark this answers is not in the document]
rgaudin: You're right, that first test is duplicated. Removed it

rgaudin:
@mgautierfr it seems the tests now fail as we are expecting a Runtime Error exception when adding content to a non-started Creator but we segfault instead. It was not the case earlier so that may have to do with your edits…

Those tests are in `test_libzim_creator.py` at `test_creator_additem()`, `test_creator_metadata()` and `test_creator_redirection()`. Additionaly, `test_creator_badfilename()` segfaults as well. Looks like the same issue.

It seems to not fail on github actions (failing ones are reader-related) but it does locally here.

Issue #116, opened 2021-08-11
Clean-up repo


rgaudin:
There are a few things that should maybe be removed or moved:

- `examples/` folder has a single example which is not up-to-date (clustersize, hints)
- `run_tests.sh`
- `rebuild.sh`
- `lib/` and `include/` folders
- `Pipfile` duplicating pyproject.toml
- #70 might require some changes to those and/or README.

I'd rather have an extended and more user-friendly README than all those files in the root of the repo.

mgautierfr:
The lib and include folders are here to allow the user/developer to put the headers and library locally and have python-libzim use them.
Do we want to remove them ?

Do we move to something like ? :
- Put the library/headers somewhere
- Be sure that your system is configured correctly to find them (system package, CFLAGS/LDFLAG/LD_LIBRARY_PATH)
- We don't care about local files

rgaudin:
- we should definitely be able to use packaged versions when present. I believe this is already covered and your PR probably suppresses the warnings reported in #70.
- most new comers will want to use non-packaged version I suppose and we should make it easy for them. Having them to configure their system to make it findable seems like too much trouble.
- current solution is fine but it's confusing ; partly because of the similarities of names between libzim and the wrapper.

I'd advise in keeping it as it is but instead of having those two folders in the root, have a placeholder like `libzim_dist` or something similar in which to extract the downloaded version ; which itself has the two folders. Clearer, easier to replace with different version.

PR #120, opened 2021-09-22
Clean up

mgautierfr:


tasks.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@task
+def install_dev(c):
+    c.run("pip install -r requirements-dev.txt")
+
+
+@task
+def lint(c):
+    c.run("isort .")
```
rgaudin, line 34: Don't you want to use `isort --profile black .` instead ?

kelson42:
DOes it fix #116

rgaudin:
Yes and should/will. it's still in draft

rgaudin:
Github actions is malfunctioning. Awaiting it to be back to merge this

PR #122, opened 2021-10-06
Remove duplicated bits 

rgaudin:
- moved flake8 config to setup.cfg to save a file
- removed --profile param on isort in tasks as setup.cfg specifies it
- removed --max-line-length param on flake8 in tasks as setup.cfg specifies it
- removed Pipfile as dev dependencies are specified in dev-requirements.txt
- updated .gitignore to new file naming and added placeholder for dev files
- Restored a search test as upstream is fixed


mgautierfr (approved):
I had already removed Pipfile localy but forget to commit the change.
I've also removed run_tests.sh as the invoke/task.py take care of this.

PR #133, opened 2022-01-10
Skip cython on setup.py clean

rgaudin:
At least for debian integration, allow setup.py clean to run without Cython
build-dependency to be installed.

If call with just that clean command, Cython conf will be skipped.
If call as setup.py clean build_ext, then Cython is not skipped.


mgautierfr (approved):

legoktm:
Thanks!

PR #135, opened 2022-01-26
python3.10

jc-louis:
- update cython 0.29.22 → 0.29.23 (first supporting 3.10)
- add test and build for 3.10

It seems like tests are not setups to run on PRs, [but they passed on my fork](https://github.com/jc-louis/python-libzim/runs/4948807987?check_suite_focus=true).


rgaudin (changes requested):
Thank you @jc-louis ; I've tested with Cython 0.29.26 (latest) and it works fine as well. Could you use that version instead as we try to periodically catch-up with Cython.

Also please rebase this all into one commit so I can merge.

rgaudin (approved):

PR #143, opened 2022-05-19
Forwarding libzim exception in addIllustration()

### LINKED ISSUE #141: Segfault adding a duplicate illustration
I believe it can be considered a *bug* that we now (libzim 7.2.1) raises a RuntimeError when adding a duplicate of Entry, Metadata or Redirection but not for Illustration. Trying to add another (or the same) illustration for the same size segfaults.

Not sure if this gets fixed in here or at libzim level though.

rgaudin:
@mgautierfr I've set `nogil` as well since `addMetadata()` has it.

Fixes #141

### DIFF
--- a/libzim/libzim.pyx
+++ b/libzim/libzim.pyx
@@ -287,7 +287,12 @@ cdef class _Creator:
     def add_illustration(self, int size: pyint, content: bytes):
         """Add a PNG illustration to Archive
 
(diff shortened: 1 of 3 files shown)


mgautierfr (approved):
Easy fix :)

And yes, nogil was needed.

PR #151, opened 2022-07-11
Remove lzma compression option

### LINKED ISSUE #150: Remove ability to compress using lzma
See https://github.com/openzim/libzim/issues/714

rgaudin:
Fixes #150

### DIFF
--- a/libzim/libzim.pyx
+++ b/libzim/libzim.pyx
@@ -190,8 +190,7 @@ class Compression(enum.Enum):
     # We don't care of the exact value. The function comp_from_int will do the right
     # conversion to zim::Compression
     none = 0
-    lzma = 1
-    zstd = 2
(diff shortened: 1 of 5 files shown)


libzim/libzim.pyx
```diff
@@ -190,7 +190,6 @@ class Compression(enum.Enum):
     # We don't care of the exact value. The function comp_from_int will do the right
     # conversion to zim::Compression
     none = 0
-    lzma = 1
     zstd = 2
```
mgautierfr, line 193: Should we change zstd value to 1 (and adapt `comp_from_int`) ?
If we keep the value 2, I'm afraid that we will think in the future that the value is meaningful and it is not.

mgautierfr (approved):

PR #152, opened 2022-08-12
Update to libzim 8

rgaudin:
Uses libzim8 release and updates tests & release worflow:
- macos-10.15 has been removed
- macos-12 has no default python3.6 anymore. Using python-versions's 3.6.15 (downloaded)
- pytest7 now requires pythonpath in pyproject.toml
- using latest minor cython 0.29.32
- Updating blank ZIM's expected size. libzim8 creates a slightly larger ZIM for this.

### DIFF
--- a/tests/test_libzim_reader.py
+++ b/tests/test_libzim_reader.py
@@ -17,7 +17,7 @@ from libzim.suggestion import SuggestionSearcher
 ZIMS_DATA = {
     "blank.zim": {
         "filename": "blank.zim",
-        "filesize": 1173,
+        "filesize": 2197,
         "new_ns": True,
         "mutlipart": False,
         "zim_uuid": None,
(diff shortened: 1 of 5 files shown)


tests/test_libzim_reader.py
```diff
@@ -17,7 +17,7 @@
 ZIMS_DATA = {
     "blank.zim": {
         "filename": "blank.zim",
-        "filesize": 1173,
+        "filesize": 2197,
         "new_ns": True,
         "mutlipart": False,
```
kelson42, line 22: !? Multipart?!

rgaudin: PRs are not the place to comment on code (test actually) not modified by such PR. We test the `is_multipart` method in tests and this above is a description of what to expect for this ZIM file (no multipart)

rgaudin:
Sorry about the noise ; had difficulties finding what made the release workflow *randomly* fail.

PR #158, opened 2023-02-28
Add `printVersions()` and `getVersions()`.

FledgeXu:
Fix: #148


rgaudin:
@FledgeXu thank you very much ; I've opened #159 with my changes… I had to fix reader tests and then I wanted to improve the API. Moving there

PR #159, opened 2023-03-01
Introducing printVersions() and getVersions()

### LINKED ISSUE #148: Provides methods for printVersions() and getVersions()

rgaudin:
This is a follow-up of #158 and fixes #148 as well.

This includes @FledgeXu's PR but improves the Python API

```
- Better docstring
- Refactored to uses module functions instead of (useless) Version class
- Changed get_versions() to return an OrderedDict for easier in-python use
- Added get_libzim_version() shortcut
- Simplified print_version's code
- Fixed print_version tests (use flush so we're sure its present when checking)
- Changed tests for get_versions and get_libzim_version
```


FledgeXu:
Thanks for your improvement!

PR #163, opened 2023-03-18
Add Python 3.12 and remove 3.6 in the CI

kelson42:


kelson42:
3.12 still not available

PR #165, opened 2023-04-20
Better build system

### LINKED ISSUE #136: Use CI build wheel
[CI build wheel](https://cibuildwheel.readthedocs.io) provides a robust GH Action to build across systems and platforms.

This would be a better maintenance strategy to use it instead of our custom action details (although it's small) but the real value would be the ease to build and publish wheels once we have Support for Windows (#91) and once we have a macos-arm64 libzim binary.

rgaudin:
This introduce two changes:

- A better setup.py build
- A change to CI & CD workflows that fixes #136 

The first one is independent but would require rewriting the tests and release workflows so given the second one is dependent on the first one, I bundled them in this PR.

## Better builds

The build process is now completely integrated. No more separate downloading of libzim or tweaking of any kind.

```sh
python3 -m build
```

The main changes here are:
- integration of the libzim binary download (with support for macOS universal) with release and nightlies support
- storing libzim headers in `include/` for compilation
- storing libzim dylib in `libzim/` for linking and inside `libzim/` in the wheel
- runpath is set/adjusted accordingly
- sdist now doesn't include libzim binary
- macOS notarization included

## Using `cibuildwheel`

[cibuildwheel](https://pypi.org/project/cibuildwheel/) is a pypa project that simplifies the creation of many per-platform wheels  ; targeting common CI envs.

With a config file (inside out `project.toml`) we can specify all our targets and it will create all the wheels in a `wheelhouse/` folder in a single command

- builds the wheel
 - cross-compiling arm64 from x86_64 on macOS
 - using docker and qemu on Linux for different archs and libc (we dont use musl for now)
 - possibly on windows as well (we don't use it yet)
 - on multiple interpreters. I disabled pypy as two tests fail on pypy. Ticket to come.
- repairs the wheel
  - inspects all libs in the wheel to decide what the minimum glibc version must be and that it matchs the building host.
  - can bundle extra dependencies inside the wheel instead of relying on dynamic linking (we don't use it, libzim being self contained already)
- tests the wheel
  - run our pytest collection on the installed wheel, ensuring it's OK

### Test workflow

Rewritten to accommodate the build simplicity but is still running tests individually on various python versions on macos and linux.

Although it's tested in cibuildwheel, those tests are cheap and more verbose so it's more valuable.

### Release workflow

Replaced with a wheels one that calls cibuildwheel.
It creates all the wheels and stores them as artifact on Github.
It also creates the sdist of course.

When run on a `release`, it uploads them all to PyPi.

**Note** that this awaits the aarch64-bionic and x86_64-bionic in nightlies to be merged


mgautierfr (approved):
LGTM

PR #169, opened 2023-07-13
Accented search

rgaudin:
Second commit is just to ensure tests are passing ; will remove (and thus test would fail) before merging.


mgautierfr (approved):

rgaudin:
FYI build complains about bionic version not being bionic-compatible. I'll investigate separately

Issue #170, opened 2023-07-13
glibc version mismatch building wheel on py37 aarch64


rgaudin:
[this run](https://github.com/openzim/python-libzim/actions/runs/5542032499/jobs/10116191493) failed with

```
 auditwheel: error: cannot repair "/tmp/cibuildwheel/built_wheel/libzim-3.1.0-cp37-cp37m-linux_aarch64.whl" to "manylinux_2_28_aarch64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
```

rgaudin:
Issue is serious as it concerns all our aarch64 wheels.

Due to [glibc-usage](https://sourceware.org/glibc/wiki/Release), we have decided to support the `manylinux_2_28` [platform](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) for Python 3.8-3.11 (3.7 is now EOL).
This means that our binary libzim (and the wrapper) should be compatible (and thus use at most) glibc 2.28.

For this reason, we are building libzim on Ubuntu bionic (glibc 2.27) and our wrapper is built on [manylinux_2_28](https://quay.io/repository/pypa/manylinux_2_28_aarch64) (glibc 2.28).

pylibzim 3.1.0 (May 1st 2023) uses libzim 8.2.0 ([release](https://download.openzim.org/release/libzim/libzim_linux-aarch64-bionic-8.2.0.tar.gz))
- was [was working](https://github.com/openzim/python-libzim/actions/runs/4852111841): auditwheel did not complain.
- produced the [correct wheels](https://pypi.org/project/libzim/3.1.0/#files). `aarch64` ones are actually `2_24` compatible (`auditwheel` renames to match the minimum satisfying version)

Rebuilding with libzim 8.2.0 works as expected but 8.2.1 doesn't.

[`auditwheel`](https://github.com/pypa/auditwheel) finds symbols requiring glibc 3.4 which is not acceptable (Debian 12+/ Ubuntu 22.04+/ Fedora 35+). This is due to the recent version of GCC (libstdc++ declaring `GLIBCXX_3.4.29`) used by [the cross-compilation toolchain](https://github.com/tttapa/docker-arm-cross-toolchain) introduced in https://github.com/kiwix/kiwix-build/pull/615

[...]

PR #171, opened 2023-07-28
Using libzim 8.2.1

rgaudin:


mgautierfr (approved):
LGTM.

Just want to mention the change in the dependencies version from `>=` to `==`.
It is a bit more than just updating version number as it fixes a specific version.
If it is really want is wanted, it is ok. But better mention for nothing than let pass a unwanted change.

rgaudin:
Sorry this a a WIP ; the bionic builds are broken. I am figuring it out first.

rgaudin:
Blocked by #170

PR #173, opened 2023-09-12
platform only checked if downloading

rgaudin:
We were incorrectly checking the platform unconditionally while this is only relevant to halt the setup when we'll download libzim binary (and there wont be a matching file to download)


legoktm (approved):
LGTM and I tested a build on Debian i386 that succeeded. Thanks :)

PR #177, opened 2023-11-08
musl support

rgaudin:
musl/alpine support is mostly in the libzim itself so this uses the latest nightly which now has musl support.
This also adds support for the *latest nightly* and fixes musl detection in the download script.


mgautierfr (approved):
It took me some time to understand you want to use the redirect setup on server side to the last nightly (which I don't use at all) but once I got that, everything make sens. 😄

Seems good to me.

rgaudin:
Actually I never used it anywhere before but was told android people do it and it's convenient for pylibzim PRs in-between libzim PRs

PR #178, opened 2023-11-08
Dependencies updates and Python 3.12 

rgaudin:
Updating all our dependencies and building targets to make sure we support and publish for 3.12
Building on macos-13 as well and bump macOS min version to 12.0 to harmonize with other projects

⚠️ awaits (and includes) `musl2` until its merged


rgaudin:
Test change is due to Cython update which now does a better job at raising actual exception

requirements-dev.txt
```diff
@@ -1,8 +1,8 @@
 isort>=5.10.1,<6.0
-black>=22.10.0,<23.0
```
mgautierfr, line 2: Is this file written by you or a tool ?

If it is you, you can specify the version with the "Compatible release clause" : `~=22.10` (https://packaging.python.org/en/latest/specifications/version-specifiers/#id4)

If not, no choice.

PR #184, opened 2023-12-16
Rely on proper libzim version 9.1.0

kelson42:
Substainable implementation of #175 (Use stable libzim release)


rgaudin (approved):
👍

rgaudin:
FYI the PR was not OK ; it didn't change the libzim version used for release ; fixing that on main before releasing

kelson42:
@rgaudin OK. Current system is error prone. Libzim version should be defined only in one place and ideally not as an env. variable.

rgaudin:
I agree. We actually have it in code and the env is here to allow changing it easily. Maybe we should remove them from the workflows.

Issue #188, opened 2024-02-08
[next-major] Remove metadata conversion from python-libzim


benoit74:
When adding metadatas, the python-libzim is doing a [conversion for some types](https://github.com/openzim/python-libzim/blob/3b8645bccf57d128507c9d4be49f0b71ba227a60/libzim/libzim.pyx#L579-L582).

This should be removed because:
- the python-libzim is meant to be only a thin wrapper around the C libzim
- not all types are handled which is confusing (for instance we missed the conversion of tags list to a string)

This would need a major release since it is a breaking change.

siddheshwar-9897:
sir is this issue still open

siddheshwar-9897:
****
As per the issue, the current implementation in python-libzim automatically converts certain types of metadata (such as datetime.date and datetime.datetime objects) when they are added. The conversion includes:

**Converting dates to a YYYY-MM-DD string and encoding in UTF-8.
Automatically encoding strings in UTF-8.
This leads to two main problems:**

**Inconsistent Type Handling:** Not all types are handled consistently, such as lists (e.g., tags) not being converted to strings.
**Unnecessary Conversion**: The conversion is not required, as libzim itself is likely capable of handling the raw metadata types.
Suggested Solution:
**I propose we remove the automatic conversion logic entirely, which would simplify the code and avoid the inconsistencies. Specifically:**

**Remove date conversion**: Stop automatically converting datetime.date and datetime.datetime objects into strings.
**Remove string encoding**: Stop automatically encoding strings into UTF-8 bytes.
Let libzim handle types directly: Allow libzim to handle the metadata content as it expects, letting users handle any necessary conversions on their own.

**This will:

Make python-libzim a more straightforward wrapper that doesn’t enforce unnecessary transformations.
Avoid confusion or issues with inconsistent conversions (like missing list-to-string handling).
Give users more flexibility and control over their metadata, ensuring they're working with the exact data format they need.
Since this would be a breaking change (as it removes the automatic conversion behavior), I suggest we plan for a major release and update the documentation accordingly to inform users of the change.**

Let me know your thoughts or if you'd like to discuss further.

Best regards,
Siddheshwar Kadm

benoit74:
I'm sorry, but I won't read your comment. This is way too long/verbose and using way too much emphasis for something as straightforward as removing few lines of code. 

Propose a PR, but please be sharp.

rgaudin:
I believe we have a PR already that's ready and awaits that next major

[...]

PR #190, opened 2024-02-19
remove metadata conversion

aryanA101a:
#188


kelson42:
@aryanA101a Any update?

rgaudin (approved):
LGTM ;

**Must NOT be merged!**. This is an API break; we'll merge once we know we're ready to publish a major version

rgaudin:
converted to draft to prevent accidental merge

benoit74:
I don't think this is sufficient: type hints should be simplified as well to explicitly mention these methods support only bytes and str values now (i.e. conversion must be done by python-scraperlib or any other "user")

rgaudin:
Thank you @benoit74 ; that's right and I forgot @aryanA101a to mention that you should not force-push after having asked for a review ; it makes (and made in that case) reviewing a real pain.

rgaudin:
@aryanA101a please update type hints as suggested

PR #197, opened 2024-09-04
Adopting-ish openZIM bootstrap

### LINKED ISSUE #195: Swicth to bootstrap-like
pylibzim cannot be completely switched to our bootstrap but we should still apply all the lint/check rules. It currently still relies on isort which contradicts ruff for instance.

rgaudin:
Getting closer to openZIM bootsrap (we build via setuptools for Cython extension)

Fixes #195

- Upgraded build and dev dependencies
- Removed setup.cfg (moved metadata to pyproject.toml)
- Replaced isort and flake8 with bootstrap's ruff
- Added config for pyright (disabled for now as libzim has no type hints)
- bootstrap conf for lint/check/test
- updated linting to match

This will be followed by a PR with type stubs


README.md
```diff
@@ -24,8 +24,9 @@ Our [PyPI wheels](https://pypi.org/project/libzim/) bundle a [recent release](ht
 - macOS for `x86_64` and `arm64`
 - GNU/Linux for `x86_64`, `armhf` and `aarch64`
 - Linux+musl for `x86_64` and `aarch64`
+- Windows for `x64`
 
-Wheels are available for both CPython and PyPy.
+Wheels are available for both CPython only  (but can be built for Pypy).
```
benoit74, line 29: extra whitespace in `only  (but` => `only (but`

benoit74: `for both CPython only` => `for CPython only`

benoit74 (approved):
LGTM, see my comments before merging (minor)

README.md
rgaudin: Indeed!

PR #199, opened 2024-09-05
Introducing Type Hints

rgaudin:
As we are producing a binary extension via Cython, the regular source-based mechanism of type discovery is not available. This makes using libzim a little less pleasant for those relying on types.

This adds type stubs that we will have to manually maintain.

I had to tweak a bit the wrapper to be able to expose the get_indexdata thing. Previously, we were relying on the fact that the Item had no `get_indexdata` method to decide whether to use the libzim auto-index feature (relied on heavily for HTML entries).

Conditional method is not really compatible with static typing.

In order to retain functionality and API, I chose to add a `get_indexdata` variable on all `Item`, set to `None`. If this variable is `None` (no action taken), then we run the auto-index. If it is not None but returns NULL, then no index (no auto neither) If it is set and returns a proper IndexData, it is used. If the variable is missing (not sub-classing `Item`), run the auto-index.

In other words, behavior stays the same but type checker have a `get_indexdata` variable to look-for.

I don't see it as an API change and thus plan on including in 3.5.0 but please let me know what you thin

**Nota**: no test behavior change. changes are only to please type-checker or to silence it in cases we were not respecting it on purpose.


README.md
```diff
@@ -149,6 +149,12 @@ with Creator("test.zim") as creator:
         creator.add_item(item2)
 ```
 
+#### Type hints
+
+`libzim` being a binary extension, there is no Python source to provide types information. We provide them as type stub files. When using `pyright`, you would normally receive a warning when importing from `libzim` as there could be discrepencies between actual sources and the (manually crafted) stub files.
+
+You can disable the warning via `reportMissingModuleSource = "none"`.
```
benoit74, line 156: Is it possible to disable the warning for all files but only a given module? Disabling this import by import is going to be painful, and disable it for all files and all modules is a bit risky.

benoit74 (approved):
LGTM, see small remark but not mandatory before merging.

README.md
rgaudin: This is just a general suggestion so that users are aware of the warning and its pyright name. When I tested with scraperlib, I just added inline ignores (actually replaced the catchall that was there)

PR #204, opened 2024-09-23
Add missing f-string in setup.py

legoktm:
Otherwise you end up with:
```
Using found library at {config.found_libzim}. Adjust CFLAGS/LDFLAGS if needed
```


rgaudin:
Sorry I saw the branch and not the PR (my mistake) and thought you were not planning on pushing it. Already pushed it

PR #206, opened 2024-09-23
Require delocate only on Windows

### LINKED ISSUE #205: Only depend on delocate on needed platforms
My reading of setup.py is that `delocate` is only needed on Windows. I think we use a platform specifier to only add the dependency on Windows? https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers

(delocate isn't packaged in Debian so I've patched out the dependency for now)

rgaudin:
Fixes #205

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,14 +14,14 @@ repos:
   rev: v0.6.3
   hooks:
   - id: ruff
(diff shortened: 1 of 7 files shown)


benoit74 (approved):
Not really an expert in wheels and delocate, but at least all this makes sense to me.

PR #209, opened 2024-10-10
Add support for Python3.13

### LINKED ISSUE #208: Python 3.13 support
Python 3.13 is the new stable ; we should support it/provide wheels for it. We've already dropped 3.8 a couple weeks ago.

rgaudin:
No code nor dependency change required but we need to build, test and publish it.

Fixes #208

### DIFF
--- a/.github/workflows/QA.yaml
+++ b/.github/workflows/QA.yaml
@@ -14,6 +14,7 @@ jobs:
       - name: Set up Python ${{ matrix.python }}
         uses: actions/setup-python@v5
         with:
+          # to update to 3.13 once it lands in GH image
           python-version: "3.12"
           architecture: x64
 
--- a/.github/workflows/QA.yaml
+++ b/.github/workflows/QA.yaml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python ${{ matrix.python }}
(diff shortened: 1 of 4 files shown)


.github/workflows/QA.yaml
```diff
@@ -14,6 +14,7 @@ jobs:
       - name: Set up Python ${{ matrix.python }}
         uses: actions/setup-python@v5
         with:
+          # to update to 3.13 once it lands in GH image
```
rgaudin, line 17: That's because ATM the GH image doesn't include Python3.13 so setting it would require a download/install that would slow down this quick step for no real value.

.github/workflows/QA.yaml
```diff
@@ -14,6 +14,7 @@ jobs:
       - name: Set up Python ${{ matrix.python }}
```
benoit74, line 14: I know I missed it in a previous review, but this is wrong, we do not have a matrix here

benoit74 (approved):
LGTM, to release asap from my PoV

.github/workflows/QA.yaml
benoit74: Makes sense

.github/workflows/QA.yaml
rgaudin: Right!

Issue #200, opened 2024-09-07
[next-major] Remove traces of CamelCase


rgaudin:
As libzim uses CamelCase, we sometimes followed the convention to stick to the pure-wrapper way.
Given case is not that important and we don't follow it everywhere (it's become minor overtime), we should just remove the remaining traces and use snake case everywhere.

It's a tiny change but obviously breaks the API so should not be merged until approaching next major release

```py
# writer
    def config_nbworkers(self, nbWorkers: int) -> Self: ...  # noqa: N803
    def set_mainpath(self, mainPath: str) -> Self: ...  # noqa: N803
    def add_redirection(
        self,
        path: str,
        title: str,
        targetPath: str,  # noqa: N803
        hints: dict[Hint, int],
    ) -> None: ...
    def add_alias(
        self,
        path: str,
        title: str,
        targetPath: str,  # noqa: N803
        hints: dict[Hint, int],
    ) -> None: ...
# suggestion
    def getEstimatedMatches(self) -> int: ...  # noqa: N802
# search
    def getEstimatedMatches(self) -> int: ...  # noqa: N802
    def getResults(self, start: int, count: int) -> SearchResultSet: ...  # noqa: N802
````

SrisharanVS:
There are also some traces of Pascal case, is it better to change them to snake case too?

rgaudin:
> There are also some traces of Pascal case, is it better to change them to snake case too?

Indeed

siddheshwar-9897:
Also sir i think removing camelcases from setup file is good

rgaudin:
@siddheshwar-9897 as you can see, there is already a PR (#212) linked to this ticket which only awaits a new major release.
If you think there are additional incorrect CamelCase usage, please expose them here.

PR #217, opened 2025-02-09
…Remove traces of CamelCase #200

siddheshwar-9897:
Hay I am siddheshwar-9897…"I remove CamelCase from setup.py i think it is good practice also according to this made changes and also remove some error from setup.py"

Please notify this my first contribution if any thing mistaken by me please notify me.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,8 @@ from Cython.Distutils.build_ext import new_build_ext as build_ext
 from setuptools import Command, Extension, setup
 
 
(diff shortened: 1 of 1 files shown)


rgaudin:
This is wrong. We follow [PEP-8](https://peps.python.org/pep-0008/#class-names). Thank you for your contribution but make sure you understand a ticket first before sending code. You can ask questions on the ticket in that regard.

PR #219, opened 2025-02-17
Using ubuntu24.04 in CI as ubuntu-20.04 is about to be removed (2025-04-01)

rgaudin:
This is assuming that Linux wheels are built on manylinux images and should not depend on the host


benoit74 (approved):
LGTM

PR #220, opened 2025-03-25
Bind `Archive.get_random_entry()` method

TheEvilSkeleton:


rgaudin:
Thank you @TheEvilSkeleton ; can you add simple unit test(s) for this?

TheEvilSkeleton:
@rgaudin can you test if the test works? Running pytest locally doesn't seem to work

PR #221, opened 2025-04-17
Use libzim 9.3.0

rgaudin:
This only bumps libzim, ~~no API update included~~. This wraps API changes (Cache Control).

~~**Note**: Windows disabled as libzim is not avail for it ATM.~~


kelson42:
New libzim 9.3.0 build has been made and we have now binaries for Windows.

PR #229, opened 2025-07-30
Add searcher add archive

kelvinhammond:


kelson42:
@kelvinhammond Can younplease add a small description to the PR? Does it fix a known/tracked issue?

rgaudin (changes requested):
@kelvinhammond thank you for your PR.
As @kelson42 wrote, it's important that non-trivial contributions are explained ; justified even in this case as there is no exiting ticket.

As a new hidden feature, expending the README example would also be welcome.

On the PR itself, it can't be merged ATM because the usage is too confusing. As you have probably noticed, `getResults()` yields entry paths alone. That's fine for single-ZIM search but that's terrible for multi-ZIM ones. How do you expect to use it? Checking each result over the two ZIMs?

So we need to change the API to work differently. This kind of discussion would have been preferred on a ticket but it can happen here.
What changes do you suggest?

kelvinhammond:
> On the PR itself, it can't be merged ATM because the usage is too confusing. As you have probably noticed, `getResults()` yields entry paths alone. That's fine for single-ZIM search but that's terrible for multi-ZIM ones. How do you expect to use it? Checking each result over the two ZIMs?
> 
> So we need to change the API to work differently. This kind of discussion would have been preferred on a ticket but it can happen here. What changes do you suggest?

Yeah, I realized this afterwords and rewrote it in Javascript to use the node-libzim library.
We can close this PR as there isn't currently a way to get access to the iterator which would tell us which zim file the result belongs to.

rgaudin:
I know the feeling 😣
The API design looks like a mistake or shortcut from the early days of the bindings. It's indeed convenient to return the path as that's the general usage but I'll open a ticket as this is an artificial limitation of the binding.

PR #233, opened 2025-11-03
Add support for new illustrations APIs in libzim 9.4.0

### LINKED ISSUE #232: Use new libzim illustration API
To allow multiple illustrations.

https://github.com/openzim/libzim/pull/997 has been merged

benoit74:
Fix #232 

To be merged after https://github.com/openzim/python-libzim/pull/234 (which is simpler / narrower scope)

Changes:
- new `illustration` module with the `IllustrationInfo` class, used by `reader` and `writer` modules
- new signatures of `reader.Archive.get_illustration_item`: get "default" and get specific sizes/scales
- new method `reader.Archive.get_illustrations_infos`
- new signature of `writer.Creator.add_illustration` : pass detailed illustration infos instead of one fixed size


benoit74:
Ready for review, wheels build still failing on ubuntu arm64 for "known reasons" but rest is in place so ready to be analyzed

tests/test_illustration_info.py
```diff
@@ -0,0 +1,183 @@
...
+
+import pytest
+
+from libzim.illustration import (  # pyright: ignore [reportMissingModuleSource]
+    IllustrationInfo,
+)
+
+
+class TestIllustrationInfo:
+    """Tests for IllustrationInfo class."""
```
rgaudin: useless

libzim/libzim.pyx
```diff
@@ -1329,19 +1483,57 @@ cdef class Archive:
             return self.c_archive.hasIllustration(size)
         return self.c_archive.hasIllustration()
 
-    def get_illustration_item(self, size: pyint = None) -> Item:
+    def get_illustration_item(self, size: pyint = None, info: IllustrationInfo = None) -> Item:
         """Get the illustration Metadata item of the archive.
 
+        Args:
+            size: Optional size of the illustration (for backward compatibility).
```
rgaudin: If optional, the default should be mentioned

rgaudin (changes requested):
Thank you @benoit74,

When we designed this on libzim, it was very clear that:
- attributes are an optional, for-the-future feature that we have no real use case for. It's still not in the spec.
- independant width and height are important now
- scale is important now

We thus requested to be able to directly add illustration **directly** using width, height, scale. That's not the case with this.
If there are libzim primitives to do it, bind them, if it's the same, add the necessary python glue.

On the reader side, it's less important but I feel like a lot is left on the hands of the libzim-user.
How do I know if 48x48 has various scales? I have to loop through all the getIllustrationInfos() results…

rgaudin:
> I feel like your python-glue is something which should be implemented in the python-scraperlib.

I don't think so but this is debatable of course.

scraperlib is not intended for many users. It is, as its name implies, scraper-oriented and quite large and have many dependencies. We can do as we please there because we are the main users and requiring explicit extra types for this is OK there. We're already enforcing a lot of stuff, because we're the ones using it.

pylibzim on the other end is to be used by hopefully many users. It has zero runtime dependency, is lightweight and lean.
I am a strong advocate of the pure-binding strategy but in this case, the libzim API is technically-oriented and not user oriented. I'd prefer the [same I asked for months ago](https://github.com/openzim/libzim/issues/937#issuecomment-2934865700) which is exposing what we are using: `width`, `height`, `scale`. If it were not for those extraAttributes that were not part of the original request, we would not have this intermediate type.

It just seems silly to push this uncomfortable change to our users but let's keep this PR like this and open a separate ticket to discuss whether we want to make an exception or not.

libzim/libzim.pyx
benoit74: It is optional in the sense that you provide either directly the size or the full info. It is anyway not consistent with add_illustration, and I prefer when there is one single arg to match libzim, so let's move to this.

libzim/libzim.pyx
```diff
@@ -762,6 +776,146 @@ writer_public_objects = [
...
+        self.c_info.scale = value
+    @property
+    def extra_attributes(self) -> Dict[str, str]:
+        """Additional attributes as key-value pairs."""
+        result = {}
+        for item in self.c_info.extraAttributes:
+            result[item.first.decode('UTF-8')] = item.second.decode('UTF-8')
+        return result
+    @extra_attributes.setter
+    def extra_attributes(self, value: Dict[str, str]):
```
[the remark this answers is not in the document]
benoit74: This is indeed an important thing to mention. I've added a test + note in docstring.

libzim/zim.pxd
```diff
@@ -74,6 +89,7 @@ cdef extern from "zim/writer/creator.h" namespace "zim::writer":
         void finishZimCreation() except + nogil
         void setMainPath(string mainPath)
         void addIllustration(unsigned int size, string content) except + nogil
+        void addIllustration(const IllustrationInfo& ii, string content) nogil except +
```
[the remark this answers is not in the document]
benoit74: Nice catch, forgot about it in some rebase probably ^^

benoit74:
Hum, I'm afraid that in fact this simple glue strategy is not that simple, because it will break the API or introduce dirty glue code for `add_illustration`.

Currently second argument to add_illustration is the content. If content can be either second or third or fourth depending on other arguments, this is either brittle (we count arguments to know where content is) or breaking (we force content to be a named argument).

Do you have suggestion about how to handle this? Should we introduce different method names? I'm puzzled

Same kind of issue with the scale in `get_illustration_item(width, height, scale)` and `get_illustration_item(size, scale)`, should we have scale first or should we named arguments or something else?

rgaudin:
Well libzim 9.4.0 did break the API (version doesn't tell because they don't respect semver) so I see no reason we wouldn't allow ourselves to.

Also, it's a binding not a pure copy which it can't be, especially since python doesn't support overloading natively. To me it would only make sense to have the content the first param and the rest afterwards. It could even be a single func with `width_or_squaresize`, `height`, `scale`.

What do you think?

PR #234, opened 2025-11-04
Move to libzim 9.4.0

benoit74:
This is a minimal PR to move to libzim 9.4.0

It does not add new stuff like the new illustrations API (I propose to do this in a distinct PR https://github.com/openzim/python-libzim/pull/233)

It adapts to libzim changes:
  - remove `Archive.dirent_lookup_cache_max_size`, does not exists anymore in libzim
  - move `Archive.cluster_cache_max_size` and `Archive.cluster_cache_current_size` to methods outside of `Archive` object: `get_cluster_cache_max_size`, `set_cluster_cache_max_size`, `get_cluster_cache_current_size` (size is now in bytes)
  - deprecate usage of `get_illustration_sizes()`


rgaudin:
I use [this link](https://sourceware.org/glibc/wiki/Release) and [this one](https://sourceware.org/glibc/wiki/Glibc%20Timeline) to check which distro version has which glibc version easily. It's not up to date unfortunately but Ubuntu jammy 22.04 has glibc 2.35.

_Regular libizim_ is now built on jammy so if we were to use it, we could only link on images that supports 2.35+. We would thus need to update our `manylinux-*-image` instructions to use `manylinux_2_39` which is very recent. That would imply:

- Only supporting Ubuntu 24.04+ (dropping 18.10, 20.04, 20.10, 22.04, 22.10)
- Only supporting Debian 13+, released a month ago (dropping 10, 11, 12)

That's why we don't use _regular libzim_ in pylibzim but a special build named _manylinux_. It's built on `manylinux_2_28` for the sole purpose of pylibzim.

I see [here](https://github.com/kiwix/container-images/blob/main/kiwix-build_ci/manylinux_builder.dockerfile#L1) though that this is an **x86_64 image** so obviously there will be issues on aarch64. You can open a ticket in kiwix-build because [this](https://github.com/kiwix/kiwix-build/blob/75cfff14d7eeff9d6975f35c54ca59960dff7d05/.github/workflows/cd.yml#L142) is a problem

rgaudin:
It's [available](https://download.openzim.org/release/libzim/libzim_linux-aarch64-manylinux-9.2.1-1.tar.gz) and should be working

```
❯ auditwheel-symbols -m 2_28 libzim_linux-aarch64-manylinux-9.4.0-1/lib64/libzim.so.9.4.0
libzim.so.9.4.0 is manylinux_2_28 compliant.
```

PR #237, opened 2025-11-04
Add support for Python 3.14, upgrade dependencies and CI actions, upgrade macOS support to 14.0+, add free-threaded CPython

### LINKED ISSUE #235: What is the strategy for CI in terms of images / python versions?
Currently CI run tests and build wheels on `macos-13`, `windows-2022` and `ubuntu-24.04`.

`macos-13` image is deprecated and we should move to at least `macos-14`.

I propose we rather update to test and build wheels and sdist on all stable versions:
- `ubuntu-24.04`
- `ubuntu-22.04`
- `macos-14`
- `macos-15` (ARM)
- `macos-15-intel`
- `windows-2022`
- `windows-2025`

Multiplied with the 5 python versions we support, it means we will run 7 * 5 = 35 jobs for the tests (compared to 15 today). But these tests are pretty quick, and I feel like it is worth to check everything runs fine on all python and OS version since this is a "very shared" lib.

For wheels and sdist we do not test all Python versions, so this is less a concern.

### LINKED ISSUE #236: Add support for Python 3.14
We need to support Python 3.14

benoit74:
Fix #236 
Fix #235 

Changes:
- Drop support of Python 3.9
- Add support for Python 3.14
- Upgrade dependencies, especially Cython 3.1.6
- Upgrade supported macOS to 14.0+ (instead of 13.0+)
- Add support for free-threaded CPython (3.13 and 3.14)
- Upgrade Github CI Actions
- Run tests on minimum supported platforms + more recent stable ones

Notes:
- I had  to drop support of Python 3.9 because (at least) coverage 7.11.0 supports only Python >= 3.10 (and 3.9 is anyway EOL)
- Upgrade to Cython 3.1+ needed special attention for Python 3.13+ ; if we agree on the approach, I will open an issue to not forget to "rollback" this once `sys.monitoring` works correctly with coverage and Cython
- I'm also going to open an issue regarding GIL and free-threaded Python ; work on this PR is conservative to continue same approach done so far, but I'm questioning this is the optimal approach

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Drop support of Python 3.9, add support for Python 3.14, upgrade dependencies (#237)
+- Maintenance upgrades (#237):
+  - Drop support of Python 3.9
+  - Add support for Python 3.14
+  - Upgrade dependencies, especially Cython 3.1.6
+  - Upgrade supported macOS to 14.0+ (instead of 13.0+)
+  - Add support for free-threaded CPython (3.13 and 3.14)
+  - Upgrade Github CI Actions
+  - Run tests on minimum supported platforms + more recent stable ones
 
 ## [3.7.0] - 2025-04-18
 


tests/test_libzim_version.py
```diff
@@ -11,14 +11,14 @@
 def test_version_print_version_with_stdout(capsys):
     print_versions()
     print("", file=sys.stdout, flush=True)
-    stdout, stderr = capsys.readouterr()
+    stdout, _stderr = capsys.readouterr()
```
rgaudin:
```suggestion
    stdout, _ = capsys.readouterr()
```

tests/test_libzim_version.py
```diff
@@ -11,14 +11,14 @@
...
     assert len(stdout) != 0
 
 
 def test_version_print_version_with_stderr(capsys):
     print_versions(sys.stderr)
     print("", file=sys.stderr, flush=True)
-    stdout, stderr = capsys.readouterr()
+    _stdout, stderr = capsys.readouterr()
```
rgaudin:
```suggestion
    _, stderr = capsys.readouterr()
```

.github/workflows/CI-wheels.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   LIBZIM_DL_VERSION: "9.3.0-1"
-  MACOSX_DEPLOYMENT_TARGET: "13.0"
+  MACOSX_DEPLOYMENT_TARGET: "14.0"
```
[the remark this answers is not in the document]
benoit74: It does work.

PR #238, opened 2025-11-04
Upgrade CI actions

benoit74:
This is a test PR to nail down issue with upgrade to Python 3.14


benoit74:
#237 works correctly now, no need for this PR

PR #245, opened 2025-12-04
enhancement: added changelog to the docs

### LINKED ISSUE #227: Add changelog to docs
Noticed the docs doesn't include the changelog which would be very handy and is common in similar projects (instead of root).

aaryansinhaa:
Fixes #227 

added changelog.md to the docs

Screenshot for your reference:
image


rgaudin (approved):
Works great ; thank you ; would you mind adding an entry in the changelog to mention that we've added the Changelog to the docs? 😀

Issue #230, opened 2025-08-05
Multiple archive searcher


rgaudin:
As highlighted by #229, Searcher's `getResults()` only yields result's entry path. while convenient for single-archive search, it prevents implementing multi-ZIM search as results would only be path strings from multiple ZIMs.

We should then implement multiple ZIM search properly by
- Binding addArche to Searcher (ref impl in #229)
- Change Searcher API so that results can be identified

HelloWorld-25:
Hello! I am Nithish Balaguru, a  Information Technology & Sophomore year student in MSEC , I'm passionate about open-source development and contributing to impactful projects. I have experience in C,Python, Java, web development & Working on the GSoC 2026 project, here’s an implementation for [Add multi-archive search support and identifiable search results] and I enjoy solving challenging problems through code. I am excited to contribute to GSoC 2026, learn from the community and help advance for Project Kiwix through meaningful contributions.

PR #248, opened 2025-12-28
Add multi-archive search support and identifiable search results

HelloWorld-25:
Issue : #230  As highlighted by https://github.com/openzim/python-libzim/pull/229, Searcher's getResults() only yields result's entry path. while convenient for single-archive search, it prevents implementing multi-ZIM search as results would only be path strings from multiple ZIMs.

We should then implement multiple ZIM search properly by

Binding addArche to Searcher (ref impl in https://github.com/openzim/python-libzim/pull/229)
Change Searcher API so that results can be identified                                                                                                                                        

--------------------------------------------------------------------------------------------------------------------------------------

 # Our Changes :

# Allowing multiple archives to be bound to a Searcher

    1. Added a _archives list to track all registered archives.

     2. Introduced Searcher.addArchive(archive: Archive) method to register additional archives.

     3. Searcher.search(query) now searches across all bound archives.

# Returning identifiable search results

        1. Introduced a new class: SearchResult containing:

                  class SearchResult:
                        archive: Archive
                        path: str

        2. Updated SearchResultSet.__iter__() to yield SearchResult objects instead of strings.

       3. Results are now unambiguous and include both archive and entry path.

# Updated type hints in search.pyi

       1. SearchResultSet.__iter__() now returns Iterator[SearchResult]

       2. Searcher.addArchive() is added to type hints

       3. Python API fully matches the new Cython implementation.

----------------------------------------------------------------------------------------------------------------------------------

# Benefits : 

  1.Enables multi-ZIM search

  2.Ensures search results are uniquely identifiable

  3.Clean, maintainable API, consistent with libzim C++ internals

  4.Future-proof for features like ranking, filtering, and deduplication across multiple archives.

----------------------------------------------------------------------------------------------------------------------------------

# Backward Compatibility :

  1.The API change from Iterator[str] → Iterator[SearchResult] is intentional to support multi-ZIM search.

  2.Users can still access the path via result.path to simplify migration.

------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/libzim/libzim.pyx
+++ b/libzim/libzim.pyx
@@ -1429,10 +1429,17 @@ reader_public_objects = [
 ]
 reader = create_module(reader_module_name, reader_module_doc, reader_public_objects)
 
+###############################################################################
+#   Search module                                                             #
+###############################################################################
 
-###############################################################################
-#   Search module                                                             #
-###############################################################################
+from __future__ import annotations
+
+from collections.abc import Iterator
+
+cimport zim
+from libzim.reader cimport Archive
+from libzim._util cimport move, preincrement
 
 search_module_name = f"{__name__}.search"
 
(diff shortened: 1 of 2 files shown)


libzim/libzim.pyx
```diff
@@ -1429,144 +1429,154 @@ reader_public_objects = [
 ]
 reader = create_module(reader_module_name, reader_module_doc, reader_public_objects)
 
-
 ###############################################################################
-#   Search module                                                             #
+#   Search module                                                             #
 ###############################################################################
 
+from __future__ import annotations
```
rgaudin: Keep imports at top of file

PR #250, opened 2026-03-23
Upgrade to libzim 9.5.1 + Cython 3.2.4 + other deps

benoit74:


rgaudin (approved):
LGTM, awaiting build result 😀

I think it would be wise to update Cython as well, as we do periodically to benefit from bug fixes and ensure we're not behind on changes.

PR #263, opened 2026-08-13
Stable ABI wheels

### LINKED ISSUE #264: Build for ABI3
At the moment we build each release for all supported Python version **at time of release**.

Instead, we could build for the [Stable ABI](https://docs.python.org/3/c-api/stable.html#limited-c-api) and have one wheel per platform instead of one per platform per python runtime and version.

The main benefit is that when a new Python version is available (beta, rc, released), one can use our existing wheel.

Another benefit is that release build time should be greatly reduced (currently more than 2 hours).

rgaudin:


benoit74:
Looks like I've fixed the thing ; problem was that filename inside the wheel must be `libzim.abi3.so` on Unix and `libzim.pyd` on Windows for other Python version to load it correctly ... And it looks like Cython own logic to compute extension filename is buggy on that aspect.

Let's discuss this live next week, because I'm really unsure I got all details correctly

